PR #75, opened 2018-04-11
fix: geo.offDetector size check error

edimetia3d:


AnderBiguri:
@edimetia3d Ah I see the mistake, its the error message that its wrong!

PR #110, opened 2019-03-08
Update README.md

pavankuppa1:


AnderBiguri:
Thanks @pavankuppa1 for the corrections! We will be soon changing the README completely, the current (work in progress) version is in branch `frontispiece`, but still has some work to do.

PR #136, opened 2019-08-01
Update Axtest.py

YushanGao:


YushanGao:
I rewrite the TIGRE a bit, enabling the Ax and Atb function to choose specific GPU to do the computation. The other codes are not changed at all. However, I found that if I made a deep copy of geometry variable 'geo', the projections are then different from one using `geo'. The Axtest.py can be used to show the behaviour.

PR #141, opened 2019-10-02
Add options that control optimization for nvcc - linux

rodrigovimieiro:
I had this error:

Tmp

Then added the optimization flag for nvcc/gcc, based on these links: [1](http://gcc.gnu.org/onlinedocs/gcc-4.8.1/gcc/Optimize-Options.html#Optimize-Options) and [2](https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#options-for-altering-compiler-linker-behavior-optimize)

Finally, I was able to compile everything:

tmp2

My system configuration is:

- Ubuntu 19.04
- CUDA 10.1
- gcc (Ubuntu 8.3.0-6ubuntu1) 8.3.0
- MATLAB 2018a


AnderBiguri:
Thanks @rodrigovimieiro ! I have been putting off this bug for a while. Its in fact CUDA 10.1 that has suddenly introduced this requirements, until now CUDA was good with the flag `-O` without number. I believe this is backwards compatible.

PR #158, opened 2020-04-04
FDK for displaced detector (Half-fan)

dyycy:
## Sorry, this work was completed in Feb, but as a new hand with GitHub, I made a mistake in opening a pull request, leaving the request pending for that long.

A FDK_half.m file is added, which is modified on FDK.m.

Major Changes:

Zero-pad proj data laterally to avoid aliasing effect.
Redefine proj and geo
Preweight proj before FBP
A demo file named demo_for_FDK_Half_based_on_d14_Offsets, is attached as well for testing. Just push F5 and run it. You can also modify geo.offDetector = [offset; 0] for double check. Since geo.sDetector(1) = 409.6, offset should be within - 204.8 to 204.8.

PS: Sorry for this late submit.
Actually, I finished coding weeks ago, and it worked well. But when I applied FDK_half to my own recon job(CBCT on a radiotherapy linac), some wired stuff showed up. It really took me a while before I finally realized the problem is the data, rather than the FDK_half itself.

### DIFF
--- a/MATLAB/Demos/FDK_half.m
+++ b/MATLAB/Demos/FDK_half.m
@@ -1,0 +1,192 @@
+function [res]=FDK_half(proj,geo,angles,varargin)
+%FDK for displaced detector
+%
+%--------------------------------------------------------------------------
+%--------------------------------------------------------------------------
+% This file is part of the TIGRE Toolbox
+% 
+% Copyright (c) 2015, University of Bath and 
+%                     CERN-European Organization for Nuclear Research
+%                     All rights reserved.
+%
+% License:            Open Source under BSD. 
+%                     See the full license at
+%                     https://github.com/CERN/TIGRE/blob/master/LICENSE
+%                     and
+%                     https://www.mathworks.com/matlabcentral/fileexchange/view_license?file_info_id=35548
+%
+% Contact:            tigre.toolbox@gmail.com
+% Codes:              https://github.com/CERN/TIGRE/
+% Coded by:           (Original FDK) Kyungsang Kim, modified by Ander Biguri 
+%                     (FDK_half) Yi Du (yi.du@hotmail.com)
+%--------------------------------------------------------------------------
+%% Parameter pasring
+[filter,parker]=parse_inputs(proj,geo,angles,varargin);
+geo=checkGeo(geo,angles);
+geo.filter=filter;
+
+%Input is data,geosize,angles
+if size(geo.offDetector,2)==1
+    offset=repmat(geo.offDetector,[1 size(angles,2)]);
+else
+    offset=geo.offDetector;
+end
+
+%% Zero-padding to avoid FFT-induced alising
+[zproj, zgeo, theta] = zeropadding(proj, geo);
+
+%% Preweighting using Wang function
+[proj_w, pre_w] = preweighting(zproj, zgeo, theta);
+
+%% Replace original proj and geo
+proj = proj_w;
+geo = zgeo;
+
+%% Weight
+proj=permute(proj,[2 1 3]);
+for ii=1:size(angles,2)
+    
+    us = ((-geo.nDetector(1)/2+0.5):1:(geo.nDetector(1)/2-0.5))*geo.dDetector(1) + offset(1,ii);
+    vs = ((-geo.nDetector(2)/2+0.5):1:(geo.nDetector(2)/2-0.5))*geo.dDetector(2) + offset(2,ii);
+    [uu,vv] = meshgrid(us,vs); %detector
+    
+    %Create weight according to each detector element
+    w = (geo.DSD(ii))./sqrt((geo.DSD(ii))^2+uu.^2 + vv.^2);
+    
+    %Multiply the weights with projection data
+    proj(:,:,ii) = proj(:,:,ii).*w';
+end
+%% Fourier transform based filtering
+proj = filtering(proj,geo,angles,parker); % Not sure if offsets are good in here
+
+%RMFIELD Remove fields from a structure array.
+geo=rmfield(geo,'filter');
+%% backproject
+res=Atb((proj),geo,angles); % Weighting is inside
+
+end
+
+function [filter, parker]=parse_inputs(proj,geo,alpha,argin)
+% Parker weighting logic control
+if(range(alpha)<1.5*pi)
+    opts =     {'filter','parker'};
+    warning("Parker weighing is applied for short scan reconstruction")
+else
+    opts = {'filter'};
+    parker = 0;
+end
+
+defaults=ones(length(opts),1);
+% Check inputs
+nVarargs = length(argin);
+if mod(nVarargs,2)
+    error('CBCT:FDK:InvalidInput','Invalid number of inputs')
+end
+
+% check if option has been passed as input
+for ii=1:2:nVarargs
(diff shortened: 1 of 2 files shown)


AnderBiguri:
No worries. Let me do a code-review, and then you can let me know if you want to fix these things or you prefer to add them myself to the code.

MATLAB/Demos/FDK_half.m
```diff
@@ -0,0 +1,192 @@
+function [res]=FDK_half(proj,geo,angles,varargin)
```
AnderBiguri, line 1: I think its better to have everything under the same function `FDK()`

MATLAB/Demos/FDK_half.m
```diff
@@ -0,0 +1,192 @@
...
+end
+
+%% Zero-padding to avoid FFT-induced alising
+[zproj, zgeo, theta] = zeropadding(proj, geo);
+
+%% Preweighting using Wang function
+[proj_w, pre_w] = preweighting(zproj, zgeo, theta);
+
+%% Replace original proj and geo
+proj = proj_w;
```
AnderBiguri, line 42: We should replace them in line 39 to save memory. Perhaps `zproj` is not needed either, replacing is safe.

MATLAB/Demos/FDK_half.m
```diff
@@ -0,0 +1,192 @@
...
+geo = zgeo;
```
AnderBiguri, line 43: Same with this.

MATLAB/Demos/demo_for_FDK_Half_based_on_d14_Offsets.m
```diff
@@ -0,0 +1,60 @@
...
+close all;
+
+%% Define Geometry
+geo=defaultGeometry('nVoxel',[256;256;256]);                     
+
+% Offsets
+%% %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+% Lets try simnple offset: The detector gets completelly displaced
+geo.offOrigin = [0;0;0];                     % Offset of image from origin   (mm)              
+geo.offDetector=[180; 0];                     % Offset of Detector            (mm)
```
AnderBiguri, line 35: This is described as half offset all along this PR, but in theory it should work for any type of offset, right? Not only half offsets. Is the code in `FDK_half()` hardcoded somehow to onyl work with half offsets? If so, could we change it to allow for any offset?

MATLAB/Demos/demo_for_FDK_Half_based_on_d14_Offsets.m
```diff
@@ -0,0 +1,60 @@
...
+angles=linspace(0,2*pi,360);
+head=headPhantom(geo.nVoxel);
+projections=Ax(head,geo,angles,'interpolated');
+
+plotImg(projections, 'Dim',3)
+
+%% lets see it
+imgFDK = FDK_half(projections,geo,angles);
+
+plotImg(imgFDK,'Dim',3)
```
AnderBiguri, line 51: you can do `plotImg(imgFDK,'dim',3,'slice',64)`

dyycy:
Comments recieved. Kind of busy this week. I will review the code first, and reply later. 
BTW: I'm not well informed of the situation in Germany, but be careful and stay at home.

AnderBiguri:
@dyycy Thanks! there is absolutely no hurry! No need to rush this. 
I am in the UK :) But indeed, careful and at home! Please do the same!

MATLAB/Demos/FDK_half.m
dyycy: Yeah, I can merge it into FDK. But my main concern is, probably that will make the code in FDK hard to read, which is not friendly to TIGRE users, especially new ones that are no familiar with CT recon.

MATLAB/Demos/demo_for_FDK_Half_based_on_d14_Offsets.m
dyycy: I agree

MATLAB/Demos/FDK_half.m
AnderBiguri: Hum, I still think its better suited for TIGRE to have it in the same function. All algorithms are unique, and ultimately this is just a weighting factor to FDK, but still FDK.  Ideally we would have a check if the detector is displaced, and a comment explaining why this is required, or simply a link to the paper.

MATLAB/Demos/demo_for_FDK_Half_based_on_d14_Offsets.m
dyycy: _half_ here or in some papers/books doesn't mean the beam angle has be to exactly 0.5* fan angle. In the code, _geo.offDetector=[**X**; 0];_   the X can be any value. The displacement is calculated and compensated in **_zeropadding_** and **_preweighting_**.

dyycy:
To make the workflow clear for you to understand, I made a very simple 1-slide PPT. Hope that will help. After that, we can decide whether to merge it into FDK (personally I prefer a standalone function).
BTW, sorry for my poor memory, but UK is much more dangerous. Be safe.
[FDK_half workflow.pptx](https://github.com/CERN/TIGRE/files/4450271/FDK_half.workflow.pptx)

MATLAB/Demos/FDK_half.m
```diff
@@ -0,0 +1,192 @@
...
+end
+
+end
+
+function [zproj, zgeo, theta] = zeropadding(proj, geo)
+% ZEROPADDING as preprocessing for preweighting
+zgeo = geo;
+
+padwidth = fix(2*geo.offDetector(1)./geo.dDetector(1));
+zgeo.offDetector(1,:) = geo.offDetector(1,:) - padwidth/2 * geo.dDetector(1);
```
AnderBiguri, line 141: This should be equivalent to `zgeo.offDetector(1,:)=0`?

AnderBiguri: Hum, maybe not due to decimal values in pixels.

MATLAB/Demos/FDK_half.m
dyycy: OK. I'm fine to merge it into FDK. But I'm kind of busy recently, it will take a while to finish the work. If you prefer doing the job to keep uniform code style, I can explain all the details of _FDK_half_ to facilitate your recoding.

AnderBiguri:
@dyycy  no hurries and no worries. I am also quite busy, so if I get the time, I will add these changes myself. No need to do it yourself if its a lot of work! Thanks for the enormous help anyway!

dyycy:
Okay, I will leave it to you.
BTW, I'm working on a TIGRE-based module to reconstruct Varian radiotherapy linac on-board CBCT. When I finish it, could you also review the code? (Maybe we can coauthor a _technical note_ on that module)

AnderBiguri:
@dyycy sounds good. Lets chat about it further via email when you are ready

dyycy:
Just close it

AnderBiguri:
@dyycy Sure! But thanks a lot, and this is still in my inmediate TODO list. Please do not delete your branch as I still will use it to add the improvements to TIGRE. I enormously appreciate your support!

Issue #104, opened 2019-01-08
Parker weigths assume increasing positive angles.


AnderBiguri:

haozhangbit:
I feel this is a minor problem. A bigger problem is that the Parker weights are totally incorrect when the angles are not starting from 0 (e.g., angles = deg2rad(100:1:320); ).  One easy way to address this issue is: In Line 37 of ParkerWeight.m, modify from current "beta=angles(ii);" to "beta=angles(ii)-angles(1);".

AnderBiguri:
@hannahshort huh! absolutely correct!

haozhangbit:
Dr. Biguri, we could address both problems by modify only few lines of the ParkerWeight function. Please find attached picture.   I just do not understand why we need to flip alpha from left to right (original line 22, new line 24 below) if angles are incremental.  I guess it is related to the geometry setting in TIGRE, e.g., incremental angles correspond to counter-clockwise rotation (or the opposite)?

![ParkerWeight](https://user-images.githubusercontent.com/53584921/78860192-68729480-79e6-11ea-92fe-d367622e92e1.PNG)

AnderBiguri:
Honestly, I dont remember why I did that @haozhangbit. I would need to re-read the paper of the Parker weights to figure that out, but it may as well be a bug.

How are your github skills? You could make a Pull Request with your proposed changes so we can actually work on code instead of on images, comment on the changes as they happen. Are you happy to do so?

haozhangbit:
Thank you, Dr. Biguri! I am new to github, but would be happy to contribute if I can.

AnderBiguri:
Fantastic @haozhangbit ! Then, in very short, you need to "fork" this repository, add the changes that you thing are necesary, "commit" and "push" them, and finally make a "pull request", so I can review them, and if we all agree, they are then they are automatically added to the main code

PR #161, opened 2020-04-13
Update ParkerWeight.m

haozhangbit:
eliminate the requirements: 1. angles must be incremental; 2. angles must start from 0

### DIFF
--- a/MATLAB/Utilities/ParkerWeight.m
+++ b/MATLAB/Utilities/ParkerWeight.m
@@ -14,13 +14,13 @@ function data=ParkerWeight(data,geo,angles,q)
 %   year={2002},
 %   publisher={American Association of Physicists in Medicine}
 % }
-if unique(geo.DSD)>1
-   warning('Parker weigths not supported for varying geo.DSD, mean(DSD) is used.') % mean(DSD) might be better than DSD(1) for varying DSD
+if length(unique(geo.DSD))>1
+   warning('Parker weigths not supported for varying geo.DSD, mean(geo.DSD) is used.') % mean(DSD) might be better than DSD(1) for varying DSD
 end
 
 alpha = atan([-geo.sDetector(1)/2+geo.dDetector(1)/2:geo.dDetector(1):geo.sDetector(1)/2-geo.dDetector(1)/2]/mean(geo.DSD));
 diff_angles = diff(angles);
-if (all(diff_angles>=0))
+if (all(diff_angles>=0)) % if angles are incremental, flip alpha in the left-right direction
    alpha=-alpha;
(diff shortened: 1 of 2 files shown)


MATLAB/Utilities/filtering.m
```diff
@@ -43,9 +44,11 @@
     
     fproj = (real(ifft(fproj)));
     
-  
-    proj(:,:,ii) = fproj(end/2-geo.nDetector(1)/2+1:end/2+geo.nDetector(1)/2,:)/2/geo.dDetector(1)*(2*pi/  size(angles,2)   )/2*(geo.DSD(ii)/geo.DSO(ii));
-    
+    if parker
+   	proj(:,:,ii) = fproj(end/2-geo.nDetector(1)/2+1:end/2+geo.nDetector(1)/2,:)/2/geo.dDetector(1)*(2*pi/  (pi/angle_step)   )/2*(geo.DSD(ii)/geo.DSO(ii));
```
AnderBiguri, line 48: Hi! Could you explain this line please? I think I understand why you are doing this, but would it not make sense to get rid of the `2*pi` somehow? Can you share some maths/paper for me to understand this?

MATLAB/Utilities/ParkerWeight.m
```diff
@@ -15,14 +15,16 @@
 %   publisher={American Association of Physicists in Medicine}
 % }
 if unique(geo.DSD)>1
```
AnderBiguri, line 17: There is a big here that was there in my code. This should be `if length(unique(geo.DSD))>1`. Could you add that?

MATLAB/Utilities/filtering.m
haozhangbit: Hi Ander, I referred to code by Kyungsang Kim (https://www.mathworks.com/matlabcentral/fileexchange/35548-3d-cone-beam-ct-cbct-projection-backprojection-fdk-iterative-reconstruction-matlab-examples), and do not fully understand the math yet. I will try to figure it out later. 

BTW, which reference did you use for the filtering weighting for non-Parker one? 
proj(:,:,ii) = fproj(end/2-geo.nDetector(1)/2+1:end/2+geo.nDetector(1)/2,:)/2/geo.dDetector(1)*(2*pi/  size(angles,2)   )/2*(geo.DSD(ii)/geo.DSO(ii));

AnderBiguri: I actually took it from there! Ah, I removed the parker apparently  (this was over 6 years ago).
My bad!

AnderBiguri:
This Resolves #104

Issue #168, opened 2020-07-08
Detector rotation not working on python parallel beam


AnderBiguri:
## Summary
[WKrauze 's modification](https://github.com/CERN/TIGRE/issues/142) worked fine for 'cone' mode.

## Expected Behavior
Works for both 'cone' and 'parallel' mode.

## Actual Behavior
Works only for 'cone' mode.
The rotation of the detector does not work in the 'parallel' mode.

## Code
```
nangles = 50
angles = np.linspace(-np.pi/2, np.pi/2, nangles, dtype=np.float32)

dRot = 45

geo = tigre.geometry(mode='parallel', default=True, nVoxel=np.array([250, 250, 250]))
geo.dDetector = np.array((0.5, 0.5))*3
geo.sDetector = geo.dDetector * geo.nDetector
geo.DSO = 500
geo.DSD = 700
geo.rotDetector = np.array((0, 0, dRot*np.pi/180))

geo2 = tigre.geometry(mode='parallel', default=True, nVoxel=np.array([250, 250, 250]))
geo2.dDetector = np.array((0.5, 0.5))*3
geo2.sDetector = geo2.dDetector * geo2.nDetector
geo2.DSO = 500
geo2.DSD = 700
geo2.rotDetector = np.array((0, 0, 0))

geo3 = tigre.geometry(mode='cone', default=True, nVoxel=np.array([250, 250, 250]))
geo3.dDetector = np.array((0.5, 0.5))*3
geo3.sDetector = geo3.dDetector * geo3.nDetector
geo3.DSO = 500
geo3.DSD = 700
geo3.rotDetector = np.array((0, 0 , dRot*np.pi/180))

geo4 = tigre.geometry(mode='cone', default=True, nVoxel=np.array([250, 250, 250]))
geo4.dDetector = np.array((0.5, 0.5))*3
geo4.sDetector = geo4.dDetector * geo4.nDetector
geo4.DSO = 500
geo4.DSD = 700
geo4.rotDetector = np.array((0, 0, 0))

head = data_loader.load_head_phantom(geo.nVoxel)
proj = tigre.Ax(head, geo, angles)
proj2 = tigre.Ax(head, geo2, angles)
proj3 = tigre.Ax(head, geo3, angles)
proj4 = tigre.Ax(head, geo4, angles)
#%%
plt.ion()
fig, axis = plt.subplots(2,2)
for ii in range(nangles):
    axis[0, 0].imshow(proj[ii])
[...]

PR #169, opened 2020-07-09
Fixed #142 partially.

### LINKED ISSUE #142: Detector rotation not working
## Expected Behavior
I would expect that after specifying detector rotation array I will see a difference in generated projections.

## Actual Behavior
Whether I specify detector rotation matrix or not there is no difference in generated projections. Checked with 'parallel' and 'cone' modes. When I plot the difference between projections calculated with and without rotated detector I get matrix of zeros.

```
nangles = 20
angles = np.linspace(-np.pi/2, np.pi/2, nangles, dtype=np.float32)

geo = tigre.geometry(mode='parallel', default=True, nVoxel=np.array([250, 250, 250]))
geo.DSO = 1000
geo.DSD = 2000
geo.rotDetector = np.array((2, 1, 1))

geo2 = tigre.geometry(mode='parallel', default=True, nVoxel=np.array([250, 250, 250]))
geo2.DSO = 1000
geo2.DSD = 2000
geo2.rotDetector = np.array((0, 0, 0))

head = data_loader.load_head_phantom(geo.nVoxel)
proj = tigre.Ax(head, geo, angles)
proj2 = tigre.Ax(head, geo2, angles)

plt.ion()
for ii in range(nangles):
    plt.imshow(proj[ii, :, :]-proj2[ii, :, :])
    plt.show()
    plt.pause(0.0001)
```

## Specifications 

  - Python version: 2.7 (tested also on 3.6)
  - OS: Linux Manjaro 18.1.1
  - CUDA version: 10.1

genusn:
* The detector rotation works on python cone beam.
 * Parallel beam issue will be treated in #168.

# Test code
```
#%%
from __future__ import print_function
from __future__ import division
import tigre
import copy
import tigre.algorithms as algs
import numpy as np
import time
import sys
from tigre.demos.Test_data import data_loader
from matplotlib import pyplot as plt

#%%

nangles = 50
angles = np.linspace(-np.pi/2, np.pi/2, nangles, dtype=np.float32)

dRoll = 10
dPitch = 0
dYaw = 0

geo1 = tigre.geometry(mode='parallel', default=True, nVoxel=np.array([250, 250, 250]))
geo1.dDetector = np.array((0.5, 0.5))*3
geo1.sDetector = geo1.dDetector * geo1.nDetector
geo1.DSO = 500
geo1.DSD = 700
geo1.rotDetector = np.array((dYaw*np.pi/180, dPitch*np.pi/180, dRoll*np.pi/180))   #  ??  , u  , u x v

geo2 = tigre.geometry(mode='parallel', default=True, nVoxel=np.array([250, 250, 250]))
geo2.dDetector = np.array((0.5, 0.5))*3
geo2.sDetector = geo2.dDetector * geo2.nDetector
geo2.DSO = 500
geo2.DSD = 700
geo2.rotDetector = np.array((0, 0, 0))

geo3 = tigre.geometry(mode='cone', default=True, nVoxel=np.array([250, 250, 250]))
geo3.dDetector = np.array((0.5, 0.5))*3
geo3.sDetector = geo3.dDetector * geo3.nDetector
geo3.DSO = 500
geo3.DSD = 700
geo3.rotDetector = np.array((dYaw*np.pi/180, dPitch*np.pi/180 , dRoll*np.pi/180))

geo4 = tigre.geometry(mode='cone', default=True, nVoxel=np.array([250, 250, 250]))
geo4.dDetector = np.array((0.5, 0.5))*3
geo4.sDetector = geo4.dDetector * geo4.nDetector
geo4.DSO = 500
geo4.DSD = 700
geo4.rotDetector = np.array((0, 0, 0))

head = data_loader.load_head_phantom(geo1.nVoxel)
proj1 = tigre.Ax(head, geo1, angles)
proj2 = tigre.Ax(head, geo2, angles)
proj3 = tigre.Ax(head, geo3, angles)
proj4 = tigre.Ax(head, geo4, angles)
#%%
vol1 = algs.fdk(proj1,geo1,angles)
vol2 = algs.fdk(proj2,geo2,angles)
vol3 = algs.fdk(proj3,geo3,angles)
vol4 = algs.fdk(proj4,geo4,angles)

#%%
plt.ion()
fig, axis = plt.subplots(2,2, figsize=(8,8))
for ii in range(nangles):
    axis[0, 0].imshow(proj1[ii])
    axis[0, 0].set_title("Parallel (R,P,Y) = ({}, {}, {}).".format(dRoll, dPitch, dYaw))
    axis[0, 1].imshow(proj2[ii])
    axis[0, 1].set_title("Parallel no rotation")
    axis[1, 0].imshow(proj3[ii])
    axis[1, 0].set_title("Cone (R,P,Y) = ({}, {}, {}).".format(dRoll, dPitch, dYaw))
    axis[1, 1].imshow(proj4[ii])
    axis[1, 1].set_title("Cone no rotation")
    # plt.show()
    plt.pause(0.0001)

# %%
iZ = vol3.shape[0]//2
fig, axis = plt.subplots(2,2, figsize=(8,8))
fig.suptitle("Axial")
axis[0, 0].imshow(vol1[iZ])
axis[0, 0].set_title("Parallel (R,P,Y) = ({}, {}, {}).".format(dRoll, dPitch, dYaw))
axis[0, 1].imshow(vol2[iZ])
axis[0, 1].set_title("Parallel no rotation")
axis[1, 0].imshow(vol3[iZ])
axis[1, 0].set_title("Cone (R,P,Y) = ({}, {}, {}).".format(dRoll, dPitch, dYaw))
axis[1, 1].imshow(vol4[iZ])
axis[1, 1].set_title("Cone no rotation")

iY = vol3.shape[1]//2
fig, axis = plt.subplots(2,2, figsize=(8,8))
fig.suptitle("Coronal")
axis[0, 0].imshow(vol1[:, iY, :])
axis[0, 0].set_title("Parallel (R,P,Y) = ({}, {}, {}).".format(dRoll, dPitch, dYaw))
axis[0, 1].imshow(vol2[:, iY, :])
axis[0, 1].set_title("Parallel no rotation")
axis[1, 0].imshow(vol3[:, iY, :])
axis[1, 0].set_title("Cone (R,P,Y) = ({}, {}, {}).".format(dRoll, dPitch, dYaw))
axis[1, 1].imshow(vol4[:, iY, :])
axis[1, 1].set_title("Cone no rotation")

iX = vol3.shape[2]//2
fig, axis = plt.subplots(2,2, figsize=(8,8))
fig.suptitle("Sagittal")
axis[0, 0].imshow(vol1[:, :, iX])
axis[0, 0].set_title("Parallel (R,P,Y) = ({}, {}, {}).".format(dRoll, dPitch, dYaw))
axis[0, 1].imshow(vol2[:, :, iX])
axis[0, 1].set_title("Parallel no rotation")
axis[1, 0].imshow(vol3[:, :, iX])
axis[1, 0].set_title("Cone (R,P,Y) = ({}, {}, {}).".format(dRoll, dPitch, dYaw))
axis[1, 1].imshow(vol4[:, :, iX])
axis[1, 1].set_title("Cone no rotation")
```

### DIFF
--- a/Python/tigre/Source/_types.pxd
+++ b/Python/tigre/Source/_types.pxd
@@ -136,9 +136,9 @@ cdef inline Geometry* convert_to_c_geometry(p_geometry, int total_projections):
     if not c_geom.dYaw:
         raise MemoryError()
     for i in range (total_projections):
-        c_geom.dRoll[i] = p_geometry.rotDetector[i][0]
+        c_geom.dRoll[i] = p_geometry.rotDetector[i][2]
         c_geom.dPitch[i] = p_geometry.rotDetector[i][1]
-        c_geom.dYaw[i] = p_geometry.rotDetector[i][2]
+        c_geom.dYaw[i] = p_geometry.rotDetector[i][0]
 
     # The base unit we are working with in mm.
     c_geom.unitX = 1
--- a/Python/tigre/Source/_types.pxd
+++ b/Python/tigre/Source/_types.pxd
@@ -136,9 +136,9 @@ cdef inline Geometry* convert_to_c_geometry(p_geometry, int total_projections):
     if not c_geom.dYaw:
         raise MemoryError()
     for i in range (total_projections):
-        c_geom.dRoll[i] = 0
-        c_geom.dPitch[i] = 0
-        c_geom.dYaw[i] = 0
+        c_geom.dRoll[i] = p_geometry.rotDetector[i][0]
+        c_geom.dPitch[i] = p_geometry.rotDetector[i][1]
+        c_geom.dYaw[i] = p_geometry.rotDetector[i][2]
 
     # The base unit we are working with in mm.
     c_geom.unitX = 1


AnderBiguri:
Fantastic! @genusn Could you test a simple extra thing? Can you do FDK on the non rotated and rotated CBCT projectors to see if the images reconstructed are very similar? This would test if this change also properly applies the backprojector rotation (should do, but just in case)....

genusn:
@AnderBiguri  Thank you for your reply. I updated the test code above.
Here is the result. If the angles are small, the reconstructed images are similar to the ones that are generated from the non-rotated projections. e.g.
Axial_R10

I expected that if the rolling angle is 90 degree, the reconstructed images are the same as non-rotated ones, but, the result was...
Axial_R90
Hmm... Is this reasonable?

AnderBiguri:
@genusn Hum this confuses me too. 
First, because the order of the rotations is the opposite of MATLAB-TIGRE, and I am not sure if its by-design (I was not the main coder here) or if I may have suggested it wrong in the other issue. In matlab, the in-plane rotation around the normal vector is `rotDetector(something,0,0)` instead of `rotDetector(0,0, something)`... Not sure if we should use the first or the second (fixing it would be only swapping the indices in the code you changed). 

However I just tested FDK in MATLAB for the same case and indeed it breaks bad. However any other algorithm seems to be doing the right thing.

AnderBiguri:
Ah, yes, It has to do with basically the definition of FDK, the filtering etc breaks bad if you get out of the standard circular trajectory. Nothing to worry about, just another reason to use iteratives, I guess.

genusn:
> Ah, yes, It has to do with basically the definition of FDK, the filtering etc breaks bad if you get out of the standard circular trajectory. Nothing to worry about, just another reason to use iteratives, I guess.

If the internal process is "filer and rotate", not "rotate and filter", then that could happen, I think.

AnderBiguri:
@genusn well, is "filter and rotate" more or less. There is no explicit rotation happening, simply the geometry is what it is. But indeed, the filtering happened first, then the filtered projections are backprojected with whichever geometry is been set. This happens to some degree on the other example too, simply in there it is not as obvious that the FDK result is incorrect.

Thanks for checking this! I'll need to think a bit further in which order we want these rotations to be inputted at (as MATLAB or as is?), but this PR is good then.

AnderBiguri:
@genusn MATLAB internal stuff also flips these, so Roll is the first input. Modified that (remember to flip your angles in your code if you use master!)

Issue #151, opened 2020-02-10
Displaced detector(large-view) reconstruction problem


dyycy:
I'm working on a project based on a medical CBCT scanner. The scanner provides a so-called _full view_ imaging function, where detector is 1/4 laterally displayed.
When I reconstructed projs by myself, the results looked kind of wired. After weeks on debugging, I realized that the demo _d14_Offsets.m_ was kind of missing leading. The toolkit is not off the shelf in using FDK to solve the _detector-offset_ problem. 

## Expected Behavior
For circular trajectory CBCT, when the detector is laterally displaced less than half of its lateral size, mathematically the central slice image exactly reconstructed, and images in other slices can be reconstructed accurately enough. This technique is also called, if my memory serves me well, _displayed detector imaging_ or _large-view imaging_.

The problem is reproducible by running the following code. It's still the d14_Offset.m. I changed the _geo.offDetector_ and added a FDK recon code block.

## Actual Behavior
A shining ring artifact shows, and streaks can be seen.
![Snipaste_2020-02-10_19-22-31](https://user-images.githubusercontent.com/7495679/74145912-df83ea80-4c3a-11ea-875b-a3528898ff31.png)
![Snipaste_2020-02-10_19-20-54](https://user-images.githubusercontent.com/7495679/74145931-ead71600-4c3a-11ea-905f-faad644d605e.png)

## Code to reproduce the problem (If applicable)

%% Define Geometry
geo=defaultGeometry('nVoxel',[128;128;128]);                     

% Offsets
%% %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
% Lets try simnple offset: The detector gets completelly displaced
geo.offOrigin = [0;0;0];                     % Offset of image from origin   (mm)              
geo.offDetector = [100; 0];                     % Offset of Detector            (mm)
%% %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
% Auxiliary 
geo.accuracy=0.5;                           % Accuracy of FWD proj          (vx/sample)

%% Load data and generate projections 
% see previous demo for explanation
angles=linspace(0,2*pi,100);
head=headPhantom(geo.nVoxel);
[...]

Issue #101, opened 2018-12-04
FDK Wang weights for offseted detector


AnderBiguri:
X-ray micro-CT with a displaced detector array-Ge Wang, Medical Physics 29, 1634 (2002); doi: 10.1118/1.1489043

PR #172, opened 2020-07-22
Add Wang weights for offset detector scans

bnel1201:
Hello Ander,

I saw this originally from issue [# 151](https://github.com/CERN/TIGRE/issues/151)  and thought I would try implementing [Ge Wang's paper](https://aapm.onlinelibrary.wiley.com/doi/epdf/10.1118/1.1489043). The following works with my projection data as well as with simulations from the demo data in d14_Offsets.m.  You mentioned in your comments to #151 that this issue is specific to FDK so I think this added control flow should be sufficient. This is my first time contributing to this project and I'm to work with you if any thing needs to be changed to better fit the project  workflow.

Other notes:
Similar to the original paper, the following works for up to 30% offset, beyond that I get some shading in the center. It's likely pixel re-binning may be necessary for offsets between 30% and the max 50%.

### DIFF
--- a/MATLAB/Algorithms/FDK.m
+++ b/MATLAB/Algorithms/FDK.m
@@ -18,12 +18,12 @@ function [res]=FDK(proj,geo,angles,varargin)
 %
 % Contact:            tigre.toolbox@gmail.com
 % Codes:              https://github.com/CERN/TIGRE/
-% Coded by:           Kyungsang Kim, modified by Ander Biguri 
+% Coded by:           Kyungsang Kim, modified by Ander Biguri, Brandon Nelson 
 %--------------------------------------------------------------------------
 [filter,parker]=parse_inputs(proj,geo,angles,varargin);
 
-if abs(geo.offDetector(1)) > 0
-    proj = apply_wang_weights(proj, geo);
+if apply_wang_weights(geo)
+    proj = wang_displaced_detector_weighting(proj, geo);
 end
 
 geo=checkGeo(geo,angles);
@@ -123,13 +123,12 @@ end
 
 end
 
-function proj = apply_wang_weights(proj, geo)
-
+function proj = wang_displaced_detector_weighting(proj, geo)
     overlap_in_mm = geo.sDetector(1)/2 - abs(geo.offDetector(1));
     overlap_in_pix = round(overlap_in_mm / geo.dDetector(1));
 
     t_in_mm = linspace(-overlap_in_mm, overlap_in_mm, 2 * overlap_in_pix);
-    R_in_mm = geo.DSO;
+    R_in_mm = geo.DSO(1);
     w = 0.5 * (sin((pi * atan(t_in_mm / R_in_mm)) / (2 * atan(overlap_in_mm / R_in_mm))) + 1);
     w = repmat(w, [geo.nDetector(1), 1, size(proj, 3)]);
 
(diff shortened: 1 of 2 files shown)


AnderBiguri:
@bnel1201 ah this is fantastically unexpected! Thanks a lot for doing this! Let me give it a try myself tomorrow and I will merge it if there is no issue (there should not be!)

I assume this resolves #151 and #101 then!

MATLAB/Algorithms/FDK.m
```diff
@@ -21,6 +21,11 @@
 % Coded by:           Kyungsang Kim, modified by Ander Biguri
```
AnderBiguri, line 21: Add your name here (if you want)

MATLAB/Algorithms/FDK.m
```diff
@@ -116,4 +121,23 @@
...
 
+end
+
+function proj = apply_wang_weights(proj, geo)
+
+    overlap_in_mm = geo.sDetector(1)/2 - abs(geo.offDetector(1));
+    overlap_in_pix = round(overlap_in_mm / geo.dDetector(1));
+
+    t_in_mm = linspace(-overlap_in_mm, overlap_in_mm, 2 * overlap_in_pix);
+    R_in_mm = geo.DSO;
```
AnderBiguri, line 132: While geometries where DSO changes per-projection should not be used with FDK, they are still supported and they may give us a bug in the future if someone tries to run this with that. I suggest changing this to 

```
R_in_mm = geo.DSO(1);
```

Not necessarily correct for changing DSOs, but neither is using FDK and Wang weights, so whatever. It solves our possible bug. maybe we can leave a comment mentioning that.

MATLAB/Algorithms/FDK.m
```diff
@@ -116,4 +121,23 @@
...
+    w = 0.5 * (sin((pi * atan(t_in_mm / R_in_mm)) / (2 * atan(overlap_in_mm / R_in_mm))) + 1);
+    w = repmat(w, [geo.nDetector(1), 1, size(proj, 3)]);
+
+    wang_weights = ones(size(proj));
+    wang_weights(:, 1:size(w, 2), :) = w;
+    if sign(geo.offDetector(1)) < 0
```
AnderBiguri, line 138: Similar to the comment above (and not only here), we need to think about what happens if someone adds a varying offDetector (per projection). Do we compute `overlap_in_mm` for each offset, or do we just use the first value? 

I genuinely don't know, so I am happy to get input in here. If we just take the first value, then maybe its worth adding a warning in the cases that there are more informing the user that only the first will be used.

AnderBiguri:
> Similar to the original paper, the following works for up to 30% offset, beyond that I get some shading in the center. It's likely pixel re-binning may be necessary for offsets between 30% and the max 50%.

maybe we can also add a warning (or open a github issue) for this, so users know that this can happen? What do you think?

MATLAB/Algorithms/FDK.m
bnel1201: I completely agree, it worked for my situation but I'll fix this so it doesn't mess up any's code down the line.

AnderBiguri: Great! Feel free to fix all these, otherwise I might find some time next week to apply them myself, if you can't do it. ;)

bnel1201:
> > Similar to the original paper, the following works for up to 30% offset, beyond that I get some shading in the center. It's likely pixel re-binning may be necessary for offsets between 30% and the max 50%.
> 
> maybe we can also add a warning (or open a github issue) for this, so users know that this can happen? What do you think?

For example this is what I get with 34% displacement.
34percent

I'll add a warning in the code if the offset exceeds 30%. But if a user gets this warning it would be nice to at least point them in the right direction so adding an issue would be a good idea as well. I have a rebinning function that I can clean up and contribute later. I know  [tomopy](https://tomopy.readthedocs.io/en/latest/api/tomopy.misc.morph.html#:~:text=misc.morph.sino_360_to_180%20%28data%2C%20overlap%20%3D%200%2C%20rotation%20%3D%20%27left%27%29,is%20odd%2C%20the%20last%20projection%20will%20be%20discarded.) has this re-binning available as sino_360_to_180 as an example we could point to.

bnel1201:
Thanks for all the help! I think the most recent commit addresses your concerns. I'm not sure how to handle the pre-weighting when detector offsets or DSOs are changing per-projection so I added checks that only when these are constant the pre-weighting is applied. If those checks aren't satisfied then the pre-weighting is skipped and the usual FDK behavior continues.

AnderBiguri:
@bnel1201 I have a couple of tiny changes extra that I would like to add (couple of warnings, nothing to functionality). Is it OK if I push a commit to `bnel1201:master`?

bnel1201:
Please do!

Issue #146, opened 2019-12-02
Create geometry class


AnderBiguri:
The geometry structure has been getting more flexible and complicated with time. Having a class, instead of an struct, to describe the geometry and add methods that apply to the geometry (such as error checking or updating) would be the best approach as software dev.

ashwinpn:
Can I work on this? Also, could you redirect me to the geometry struct code?

AnderBiguri:
Hi @ashwinpn ,

Yes, that is great! Please do a fork, then create a new branch and work there so you can do a Pull request when you think its finished. 

The thing is that there is not a "geometry struct code", its simply defined each time you want a script. You can see it here, for example:

https://github.com/CERN/TIGRE/blob/master/MATLAB/Demos/d01_CreateGeometry.m#L69-L108

But its there in all other demos. The idea would be to create a class that can then have different member methods to be initialized, or to modify it, such as `checkGeometry()` that already exist as a separate function. 
The issue is, of course, that this structure is used in *almost all* functions in TIGRE, so creating a class would also probably mean to changing almost all functions in TIGRE.... Maybe its not necessary, but it may be.

ashwinpn:
Thanks! I will check it out.

PR #173, opened 2020-07-25
Geometryclass

bnel1201:
Hi @AnderBiguri  and @ashwinpn,

Is there still interest in this from [#146](https://github.com/CERN/TIGRE/issues/146)? I've added an interface which sets an agreement on the required fields and then new geometries can be subclassed from that. I turned defaultGeometry, which was a function before, into an example subclass. This doesn't change any functionality, and didnt break in going through the demos or my code, but it opens up the possibility to have different behavior depending on the geometry type. This can be helpful in removing if-else statements.


AnderBiguri:
Ah this is great! If you are interested, this is indeed something that can lead to a lot of code removal in TIGRE (which is great!).

Sounds like a good thing, and I haven't really heard from @ashwinpn, so its good if you want to tackle this. 
Just a word of advise, I can definetly test and review, advise, guide, comment and test on this, but I won't be able to get my hands very dirty with the code, so if you want to proper tackle this, then it will be on you ;) 

If you are cool with that, then this can be a great idea. It can help remove all the ugly aux variables such as the "alpha_blocks" and indexes on SART type minimization (e.g. https://github.com/CERN/TIGRE/blob/master/MATLAB/Algorithms/OS_SART.m#L86)  or all the ugly if chunks like https://github.com/CERN/TIGRE/blob/master/MATLAB/Algorithms/OS_SART.m#L136-L152. Do let me know if you want to tackle it and if you want any advise on where to start! 

And please feel free to send me an email if thre anything to discuss that may be too long for Github (I am happy to keep the conversations here too :) )

bnel1201:
Sounds good! I'll push forward on this and will let you know when I need assistance.

AnderBiguri:
I will go adding TODOs into https://github.com/CERN/TIGRE/projects/3

bnel1201:
Very good. For the first card, the short-term solution is to spit out class properties into a struct so as to be compatible with the MATLAB C API. I'll bet the C++ API can handle classes but that would be more of a rewrite and people would have to recompile.

I'll start on the MATLAB side, using more OOP language features can clean up those variables you mentioned. For more specific details I'll follow up with email but I wanted to put the game plan on GitHub.

AnderBiguri:
@bnel1201 sounds good! Feel free to add anything to the project, just created them so we can keep better track of what may be missing :)

AnderBiguri:
This looks dead now. I will close it, but if anyone is interested in reviving it, let me know and I will reopen it :) 

Thanks for the work anyway!

Issue #152, opened 2020-02-26
Multi-GPU python


AnderBiguri:
Here I will explain what I think its needed to make the multi-GPU version of TIGRE work in python. As I don't currently have access to a 2 GPU machine, I can not test these, so I will leave it open until I (or someone else) can test these.

### Main differences between python and MATLAB source code

The main issue between the two is that MATLAB stores matrices in column mayor (Fortran style) and python has them in row mayor (C styles). This means that when the 3D arrays are unrolled, `img[1]` in the source code will have a different value. However, python is now using basically the same source code as MATLAB does, which means it requires permuting the input and output of the source code in CUDA to make it work. This is a massive waste of memory, as it requires a full copy of the object, and permuting is not a cheap operation. This is the main reason why the multi-GPU code is not implemented in python, as we wanted to fix this first, which we never did.... 

This is reflected in the code in the projection CUDA calling function `_Ax.pyx` in https://github.com/CERN/TIGRE/blob/45856141b2ddd918bcd5825fe3d0cc5aa104c01c/Python/tigre/Source/_Ax.pyx#L65 and in https://github.com/CERN/TIGRE/blob/45856141b2ddd918bcd5825fe3d0cc5aa104c01c/Python/tigre/Source/_Ax.pyx#L94 
Similar lines can be found in `_Atb.pyx`. In order to make the Multi-GPU code work, these need to go, and the computation order of the pixels on the GPU kernels need to change.  

Other minor differences between source code:

- All `mexPrintf` should be changed to the standard `printf`

- The error checking code is defined with the C++ `inline` way https://github.com/CERN/TIGRE/blob/45856141b2ddd918bcd5825fe3d0cc5aa104c01c/Python/tigre/Source/Siddon_projection.cu#L58 instead of the C `#define` macro https://github.com/CERN/TIGRE/blob/45856141b2ddd918bcd5825fe3d0cc5aa104c01c/MATLAB/Source/Siddon_projection.cu#L57

- All `cudaCheckErrors` now return `int` instead of raising an error, so they need to be surrounded by an if clause. e.g. https://github.com/CERN/TIGRE/blob/45856141b2ddd918bcd5825fe3d0cc5aa104c01c/Python/tigre/Source/Siddon_projection.cu#L253 vs https://github.com/CERN/TIGRE/blob/45856141b2ddd918bcd5825fe3d0cc5aa104c01c/MATLAB/Source/Siddon_projection.cu#L332 

### What needs changing

[...]

PR #184, opened 2020-10-20
Fix memory leak

genusn:
## Summary
I found code that was suspected of having a memory leak. I propose a solution.

## Background
 * I am preparing a PR for #152. Because it will be based on the code for MATLAB, I want to get rid of my concerns.
 * No memory leak has been detected, but it should have occurred.

### Code
```
clear;
close all;
geo=defaultGeometry();
angles=linspace(0,2*pi,100);
head=headPhantom(geo.nVoxel);

plotProj(projections,angles)

projections=Ax(head,geo,angles,'ray-voxel');
plotProj(projections,angles)

geo.mode='parallel';
plotProj(projections,angles)

projections=Ax(head,geo,angles,'ray-voxel');
plotProj(projections,angles)
```

### Environment
|**Windows**| 10 |
|**MATLAB**|  2016a |
|**CUDA**| 10.1|
|**Visual Studio**| 2015|
|**GPU**| GTX1060 |

|**Windows**| 10 |
|**MATLAB**|  2016a |
|**CUDA**| 10.1|
|**Visual Studio**| 2015|
|**GPU**| 2 x GTX2080Ti |

### Result 
 * It seems that it is working.


AnderBiguri:
Thanks. I'll check how well this runs, as its much tidier code. 

However there is a reason why there is no memory leak: you have freed arrays that are just few elements long, so they just take tens of bytes at worst. On top of that, when MATLAB exits a `mex` file, it cleans up everything, essentially the call terminates, as if it was a separate executable (in essence it is). So calling the same function repeatedly will not use more memory, simply because its freed automatically at termination. This is why only the code for GPU memory is really freed in the CUDA code. 

Now, I 100% agree one has to be tidy and free memory properly!

AnderBiguri:
Thanks. I'll check how well this runs, as its much tidier code. 

However there is a reason why there is no memory leak: you have freed arrays that are just few elements long, so they just take tens of bytes at worst. On top of that, when MATLAB exits a `mex` file, it cleans up everything, essentially the call terminates, as if it was a separate executable (in essence it is). So calling the same function repeatedly will not use more memory, simply because its freed automatically at termination. This is why only the code for GPU memory is really freed in the CUDA code. 
I think python works the same


Now, I 100% agree one has to be tidy and free memory properly!

Issue #132, opened 2019-07-18
SART crashes when running on more than 2 GPU nodes


xliu68:
## Expected Behavior
I was running SART recon on a HPC cluster (CentOS system). I can specify the number of CPU and GPU nodes that I use.

## Actual Behavior
When I run SART with more than 2 GPU nodes (e.g. 4 or 8), MATLAB crashes and gives me either a "segmentation fault" (invalid next size) or a "double free" error. It also shows "Picked up _JAVA_OPTIONS: -Xmx256m". This has not been observed with FDK and OS-SART. Reducing the number of GPU nodes to 2 avoids the crashes for now. I am not sure whether this behavior is expected.

## Code to reproduce the problem (If applicable)

```
geo=defaultGeometry('nVoxel',[128;128;128]);                     
angles=linspace(0,2*pi,100);
head=headPhantom(geo.nVoxel);
projections=Ax(head,geo,angles,'interpolated');
img = SART(projections,geo,angles,1);
```

## Specifications 

  - MATLAB version: 2019a
  - OS: CentOS 7.3
  - CUDA version: 9.2.148.1

AnderBiguri:
@xliu68 Thanks for the report!  I have limited access to multi-GPU machines with more than 2 nodes, I will however put this on my priority list.  Does this only happen with SART?

xliu68:
I have only observed this in SART so far. Haven't tested in all algorithms.

bnel1201:
I was experiencing this same issue as well with my 4-gpu box and I think I can narrow down the problem a bit more. The issue seems to be with the forward projector Ax.m and the corresponding cuda calls to `interpolation_projection` in `ray_interpolated_projection.cu`. I get a memory access error anytime I try to forward-project a number angles of less than my GPU count - 1. I'm not sure where the -1 comes from, with my 4-gpu setup I get access errors when forward project 1 and 2 angles, but not with 3 or more.

SART brings this problem up because it forward projects a single projection angle at a time, so my work-around for my multi-gpu case is to use OS-SART where a batch of angles are used.

The code below reproduces the source of the problem for my 4-gpu box:
angles=linspace(0,2*pi, gpuDeviceCount-2);
head=headPhantom(geo.nVoxel);
projections=Ax(head,geo,angles,'interpolated');

Some solutions I could see would be adding a warning to SART if gpuDeviceCount > 2 and pointing the user to OS_SART

AnderBiguri:
[...]

PR #186, opened 2020-10-25
Eliminated potential double-free

genusn:
## Summary
I found code that has potential double free. This can be a problem in the migration from MATLAB to Python. I propose a solution.

## Background
 * I am preparing a PR for #152.
 * This modification is necessary for Python, but it seems that the problem stems from the original MATLAB code.

## Details
 * Although geoArray[i].offOrigZ was allocated if splits>1 and not if splits==1, freeGeoArray was always freeing it.
 * As a result, offOrigZ of geo that was passed by Ax_mex.cpp is freeed by freeGeoArray if splits == 1, and not freeed if splits>0.
     * This is not an issue in MATLAB code because Ax_mex.cpp does not free geo.offOrigZ.
     * but it is an issue in Python code because free_c_geometry of _types.pxd frees geo.offOrigZ and it causes double free if splits==1.

## Discussion
 * There could be several solutions:
     1. Modify to always allocate geoArray[i].offOrigZ. (This PR)
         * Pros: There was similar part in voxel_backprojection.cu.
         * Cons: Waste of memory (but its only 4 x nangles bytes)
     2. Remove free(geoArray[i].offOrigZ) from free_c_geometry
         * Pros: No modification necessary in MATLAB code.
         * Cons: The life of memory is different depending on conditions.
     3. Add more code to Geometry class to manage the memory by itself
         * Pros: Readability for some of developers
         * Cons: Affects all existing cpp and cu files.

### Raw Ax
```
clear;
close all;
geo=defaultGeometry('nVoxel',[128;128;128]*2);     
angles=linspace(0,2*pi,100);
head=headPhantom(geo.nVoxel);

plotProj(projections,angles)

projections=Ax(head,geo,angles,'ray-voxel');
plotProj(projections,angles)
```
In the most of the environment, the splits will be 1.
In order to check the case of splits>1, I modified a constant in checkFreeMemory from 0.95 to 0.01 and recompiled TIGRE.
(I also added a printf to check the value of splits.)

 * Both in the case of splits ==1 and splits == 2, it seemed OK.

### Reconstruction
 * d04_SimpleReconstruction.m
 * It seemed OK for splits ==1 and splits == 2.

### Environment
|**Windows**| 10 |
|**MATLAB**|  2016a |
|**CUDA**| 10.1|
|**Visual Studio**| 2015|
|**GPU**| GTX1060 |

### DIFF
--- a/MATLAB/Source/Siddon_projection.cu
+++ b/MATLAB/Source/Siddon_projection.cu
@@ -297,27 +297,17 @@ int siddon_ray_projection(float  *  img, Geometry geo, float** result,float cons
     size_t mem_proj=                  (unsigned long long)geo.nDetecU*(unsigned long long)geo.nDetecV*sizeof(float);
     
     // Does everything fit in the GPUs?
-    bool fits_in_memory=false;
+    const bool fits_in_memory = mem_image+2*PROJ_PER_BLOCK*mem_proj<mem_GPU_global;
     unsigned int splits=1;
-    Geometry * geoArray;
-    
-    
-    if (mem_image+2*PROJ_PER_BLOCK*mem_proj<mem_GPU_global){// yes it does
-        fits_in_memory=true;
-        geoArray=(Geometry*)malloc(sizeof(Geometry));
-        geoArray[0]=geo;
-    }
-    else{// Nope nope.
-        fits_in_memory=false; // Oh dear.
+    if (!fits_in_memory) {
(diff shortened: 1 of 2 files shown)


AnderBiguri:
Very interesting. This could even be the cause of the latest failures of #132, I think... I need to have a further look and try to that, but its plausible!

PR #205, opened 2020-12-02
small bug fix

yliu88au:
Fix a small bug in minimizeAwTV.m

### DIFF
--- a/ASD_POCS.m
+++ b/ASD_POCS.m
@@ -1,0 +1,378 @@
+function [ f,qualMeasOut ] = ASD_POCS(proj,geo,angles,maxiter,varargin)
+%ASD_POCS Solves the ASD_POCS total variation constrained image in 3D
+% tomography.
+%
+%   ASD_POCS(PROJ,GEO,ALPHA,NITER) solves the reconstruction problem
+%   using the projection data PROJ taken over ALPHA angles, corresponding
+%   to the geometry descrived in GEO, using NITER iterations.
+%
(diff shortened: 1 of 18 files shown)


AnderBiguri:
Hi! Thanks, this looks good. 

The problem is that the file is in a different location, thus if I "merge" it now, we will have duplicate files. 
You can try to change that in your local files, or otherwise, I will do it, no worries :)

AnderBiguri:
You have now added all MATLAB files into the main folder, which would duplicate all the files. 
(take your time to fix these things if you want, git/github is hard if you are not used to it :) )

yliu88au:
Yes, I am totally confused!

AnderBiguri:
I will cancel this pull request, so you have time to tidy. 

The general idea is to make a branch (like in the other PR) and within your computer, change your local "master" to the branch, and then just modify the files on place, and commit them.

If this is a headache, just send me the files, and I'll try to do it :)

yliu88au:
You can do it. 😊 There is a small mod on FDK to remove toolbox dependency (range()).

Yes, for MLEM default initial image should be ones. I am still totally confused about github.

Regards,
Yi

AnderBiguri:
Hi Yi,

This is the PR #205, However, you added those things you mention into PR #206, which is actually really good and tidy. 
I'd be very happy to help tidy that one up if you want, but that will mean that I will change your Github repository, to match TIGRE and tidy everything there before merging it to TIGRE.

Is it OK if I do that? I don't want to modify your code without permission

yliu88au:
Yes, go ahead please.

PR #206, opened 2020-12-02
Large MATLAB algorithms tidying, Addition of OS TV algorithms, Modification of FISTA, Python MLEM and few bug fixes

yliu88au:
Tidy up algorithms in MATLAB

### DIFF
--- a/MATLAB/Algorithms/FDK.m
+++ b/MATLAB/Algorithms/FDK.m
@@ -215,7 +215,6 @@ for ii=1:length(opts)
     switch opt
         case 'parker'
             if default
-%                parker=range(angles)<(2*pi-max(diff(angles)));
                 parker=max(angles)-min(angles)<(2*pi-max(diff(angles)));
             else
                 parker=val;
@@ -270,4 +269,3 @@ function bool = apply_wang_weights(geo)
     
     bool = true;
 end
-
--- a/MATLAB/Algorithms/FDK.m
+++ b/MATLAB/Algorithms/FDK.m
@@ -215,7 +215,8 @@ for ii=1:length(opts)
     switch opt
         case 'parker'
             if default
-                parker=range(angles)<(2*pi-max(diff(angles)));
+%                parker=range(angles)<(2*pi-max(diff(angles)));
+                parker=max(angles)-min(angles)<(2*pi-max(diff(angles)));
             else
                 parker=val;
             end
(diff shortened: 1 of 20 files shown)


AnderBiguri (changes requested):
I have added few review to the code. 
Some of it you can directly "aprove" if you like, and some of it its discussion or optional comments. 
Note that changes have to happen/will happen in your side.

MATLAB/Algorithms/FDK.m
```diff
@@ -215,7 +215,8 @@
     switch opt
         case 'parker'
             if default
-                parker=range(angles)<(2*pi-max(diff(angles)));
+%                parker=range(angles)<(2*pi-max(diff(angles)));
```
AnderBiguri, line 218:
```suggestion
```

AnderBiguri:
Note to self:

This go check in this PR before merging

- [ ] FISTA_mod maths, and see if we can add it to just FISTA
- [ ] FISTA seems changed totally, but not practically, check thaw github is doing (EOL issues?)
- [ ] Test MLEM in python

AnderBiguri:
@yliu88au , is FISTA_mod the same algorithm as in the paper that is linked in the comments? That paper has several FISTA version, is this one exactly the same as the paper calls "FISTA_mod" ?

yliu88au:
It is the version of “Lazy-Start”…..

yliu88au:
Perhaps you can make it more general to allow user to set parameter… standard FISTA, “LAZY-START”, etc….

By the way, I just mod and test Python version on newest GPU RTX3090 with CUDA version 11.1, it seems working. But my version of MATLAB does not support this GPU, so I cannot test.

AnderBiguri:
Yeah, that is a good idea. I just need to find the time to go through the code for that. 
I may add it as is and then open issues to tackle later....

Good that you could test in that new GPU!
MATLAB should not care, as unless you changed the compiling to use `mexcuda`, the standard compiler just uses whichever CUDA version you have to compile. You may need to change some flags in the `xml` file, but otherwise it should be able to compile

yliu88au:
Yeah, I modified xml file and compiled on MATLAB, unfortunately for program to work on new GPU, we do need CUDA 11.1.

AnderBiguri:
Yes, of course, but you can install CUDA 11.1 and compile TIGRE-MATLAB using CUDA 11.1 without any issue, that is what I meant.

PR #214, opened 2020-12-08
New GPU support (MATLAB)

yliu88au:


AnderBiguri:
Thanks @yliu88au !

I extended the support to python and linux/mac MATLAB

yliu88au:
Hi Ander,

I suspect that “Ax_mex.cpp” has been changed recently, it seems when I call Ax(“Siddon”), or without specify projection type, it uses “interpolation” version instead….

Regards,
Yi

AnderBiguri:
Hi @yliu88au 

Yes I see. But @genusn is trying to fix that in PR #215 , where you left a comment mentioning the issue again. I was wondering if it meant that after @genusn fix, you still see that problem, or you were just trying to make us aware :) 

Trying to fix that soon! 

Ander

Issue #208, opened 2020-12-03
2D geometry needs 0.5 vDetectorOffset??


AnderBiguri:
Investigate this.

AnderBiguri:
Code from #207 : 

```
%% Init 
clear;
close all;
%% Test parameters
test_case = 1

if test_case == 0
    % Both even
    nVoxelZ=10;
    nDetectorV=10;
elseif test_case == 1
    % nVoxel Even, nDetec Odd
    nVoxelZ=10;
    nDetectorV=11;
elseif test_case == 2
    % nVoxel Odd, nDetec Even
    nVoxelZ=11;
    nDetectorV=10;
else
    % nVoxel Odd, nDetec Odd
    nVoxelZ=11;
    nDetectorV=11;
end

nVoxelX = nVoxelZ;
nVoxelY = nVoxelZ;
nDetectorU = nDetectorV;

nangles=10;

mode='cone';  % 'parallel'
ptype= 'Siddon'; % 'interpolated' % 'ray-voxel'
%%
geo=defaultGeometry('mode',mode);
geo.DSD = 2000;
geo.DSO = 1000;
geo.nVoxel = [nVoxelX;nVoxelY;nVoxelZ];
geo.dVoxel = [1.0; 1.0; 1.0];
geo.sVoxel = geo.nVoxel .* geo.dVoxel;
geo.dDetector = [2;2] ;              % size of each pixel            (mm)
geo.nDetector = [nDetectorU; nDetectorV];     % has to be true 3D for python version
geo.sDetector = geo.dDetector .* geo.nDetector;
geo.offOrigin = [0;0;0];
geo.offDetector = [0.0; 0.0] ;
geo.rotDetector = [0; 0; 0];
geo.accuracy=0.01;

%%
angles = linspace(0, 2 * pi - 2*pi/nangles, nangles);

angles=[0,pi/2,pi,3/2*pi]
wire = zeros(geo.nVoxel', 'single');
wire(int32((geo.nVoxel(1))/2):int32((geo.nVoxel(1)+1)/2), int32((geo.nVoxel(2))/2):int32((geo.nVoxel(2)+1)/2), :) = 1;
%%
proj = Ax(wire,geo,angles, ptype);

plotProj(proj, angles,'slice',1)
```

Shows that for  test_case == 1 the projections are not symmetric. This is likely related to why #207 does not work well. 
I have not been able to reproduce for another test case, this is likely related to some bug at `compute_deltas` functions.

[...]

Issue #207, opened 2020-12-03
Python fails at 2D tomography


AnderBiguri:
The following code will return projection data that are all zeros. 

```
from __future__ import division, print_function
from __future__ import division

import tigre
import tigre.algorithms as algs
import numpy as np
from tigre.demos.Test_data import data_loader
from tigre.utilities.Measure_Quality import Measure_Quality
from matplotlib import pyplot as plt

#geo1 = tigre.geometry(mode='cone', high_quality=False, default=True)
geo = tigre.geometry(mode='cone')
geo.DSD = 1535
geo.DSO = 1000
geo.nVoxel = np.array([1, 256, 256])
geo.sVoxel = geo.nVoxel * 1.0
geo.dVoxel = geo.nVoxel / geo.sVoxel
geo.dDetector = np.array([0.8, 0.8])*2               # size of each pixel            (mm)
geo.nDetector = np.array([1, 256])     # has to be true 3D for python version
geo.sDetector = geo.dDetector * geo.nDetector
geo.offOrigin = np.array([0, 0, 0])
geo.offDetector = np.array([0.0, 0.0]) # an known bug for v1.5
geo.rotDetector = np.array([0, 0, 0])
geo.accuracy=0.5



niter = 20
nangles = 100
angles = np.linspace(0, 2 * np.pi, nangles, dtype=np.float32)
head = data_loader.load_head_phantom(geo.nVoxel)
# test_data = scipy.io.loadmat('head.mat')
# data = test_data['img'].copy()
#head = data[:,:,51:53].transpose(2,1,0).copy()
#head = np.expand_dims(data[100,:,:].copy(),axis=0)

proj = tigre.Ax(head,geo,angles)

plt.figure()
plt.imshow(proj[50,:,:])
plt.show()
exit()
```

Debugging test made:

- Changing `geo.nVoxel` to `geo.nVoxel = np.array([4, 256, 256])` will make the projections be non-zero.

- I was worried that offsets may be wrong, but changing `geo.nDetector = np.array([10,1])` shows that the image is well aligned to the center of the detector.

genusn:
@AnderBiguri , Something is happening in siddon.
```
proj = tigre.Ax(head,geo,angles, krylov="interpolated")
valMin = np.min(proj)
valMax = np.max(proj)
print(valMin, valMax)
```
showed
```
0.0 80.21488
```

AnderBiguri:
[...]

PR #218, opened 2020-12-09
Axis-parallel rays issue in Siddon

AnderBiguri:
This seems to solve it. 

- [x] more testing
- [x] Evaluate computational cost of the addition


AnderBiguri:
Seems to work well at odd detector and `angles=[0,pi/2,pi,3/2*pi]`

AnderBiguri:
Seems that its faster now that it was before, by around 4%. 

Confusing. Maybe checking all the `if` condition and failing them all (see [here](https://github.com/CERN/TIGRE/issues/207#issuecomment-741728041)) takes longer due to branching.

AnderBiguri:
This still does not work. There is a case that I found that I still don't understand:
```

%% Init 
clear;
close all;
%% Test parameters
test_case = 1

if test_case == 0
    % Both even
    nVoxelZ=10;
    nDetectorV=10;
elseif test_case == 1
    % nVoxel Even, nDetec Odd
    nVoxelZ=10;
    nDetectorV=11;
elseif test_case == 2
    % nVoxel Odd, nDetec Even
    nVoxelZ=11;
    nDetectorV=10;
else
    % nVoxel Odd, nDetec Odd
    nVoxelZ=11;
    nDetectorV=11;
end

nVoxelX = nVoxelZ;
nVoxelY = nVoxelZ;
nDetectorU = nDetectorV;

nangles=10;

mode='cone';  % 'parallel'
ptype= 'Siddon'; % 'interpolated' % 'ray-voxel'
%%
geo=defaultGeometry('mode',mode);
geo.DSD = 2000;
geo.DSO = 1000;
geo.nVoxel = [nVoxelX;nVoxelY;1];
geo.dVoxel = [1.0; 1.0; 1.0];
geo.sVoxel = geo.nVoxel .* geo.dVoxel;
geo.dDetector = [2;2] ;              % size of each pixel            (mm)
geo.nDetector = [nDetectorU; 1];     % has to be true 3D for python version
geo.sDetector = geo.dDetector .* geo.nDetector;
geo.offOrigin = [0;0;0];
geo.offDetector = -[0.,0.]' ;
geo.rotDetector = [0; 0; 0];
geo.accuracy=0.01;

%%
angles = linspace(0, 2 * pi - 2*pi/nangles, nangles);

dot = zeros(geo.nVoxel', 'single');
dot(int32((geo.nVoxel(1))/2):int32((geo.nVoxel(1)+1)/2), int32((geo.nVoxel(2))/2):int32((geo.nVoxel(2)+1)/2), int32((geo.nVoxel(3))/2)) = 1;
%%

proj = Ax(dot,geo,angles, ptype);
plotProj(proj, angles,'slice',6)
proj(:,:,6)

```

this will print 
`         0         0         0         0    0.9997         0    1.0002         0         0         0         0`
which makes no sense (there is a small tilt because the angle is not numerically exactly pi, but the middle zero makes no sense. 

If instead we use 
`dot = ones(geo.nVoxel', 'single');`
then it will print
`    4.9999   10.0002   10.0002   10.0002   10.0002   10.0002   10.0002   10.0002   10.0002   10.0002    5.0001`

The first test seem to suggest that the forward projector is missing an entire pixel in one ray (but not in the adjacent rays!). This suggest numerical issues computing that ray. However, the second test suddenly takes into account that voxel, otherwise the middle number would be missing around `1` unit in value (e.g. it would be `9.00002`)

So what the heck is going on.

AnderBiguri:
Interestingly the above only happens in this very particular geometry, any eps out of that fixes it...

genusn:
> Interestingly the above only happens in this very particular geometry

I also observed it. Under investigation...

AnderBiguri:
> > Interestingly the above only happens in this very particular geometry
> 
> I also observed it. Under investigation...

My current stage:

I see that the intersection with the `y` axis:https://github.com/CERN/TIGRE/blob/f86e18ab27aa9cb0c2d179a08552b54cd471748d/MATLAB/Source/Siddon_projection.cu#L242
happens much earlier for that pixel. Backtracking numerically on why....

AnderBiguri:
Ah, I think its a numerical precision error.

I tracked numerically the issue. We are talking about a pixel that is essentially at a axis-parallel direction, but just offseted by an incredibly small mount, due to angle being almost `pi`, but not fully `pi`. 

So, in theory, `ay=Inf` (first intersection with the `y` face of voxels). Note that I removed the eps in the denominator. 

https://github.com/CERN/TIGRE/blob/f86e18ab27aa9cb0c2d179a08552b54cd471748d/MATLAB/Source/Siddon_projection.cu#L205

For that pixel, `jmin-source.y` is `8.72612e-05` and `0.0001749992`. This shoudl be `0/0` but its not. This division precision is not good enough thus the result is a bit skewed, enough to generate a low number for `ay` (`0.4986`) and screw the line integral. Ugh.

AnderBiguri:
OK, I think this one fixes it, but I am still worried that it will kick in some other situation....

I have tested it with minimal changes to geometry, and it seems that even very small changes to `geo` generate measurable changes in the projection (i.e. this latest commit only suppresses very very very small numerical non-zero issues) and I can't seem to reproduce/find a pixel that will not be computed. I

I will leave this open for a bit, as I need to test more systematically.

AnderBiguri:
Resolve #208, Resolve #207

AnderBiguri:
@yliu88au Hi there! 

I am writing this on the pull request #218 in TIGRE (so even if you get an email, this is a public post on this particular topic). 
I was wondering if you could test this branch of TIGRE (https://github.com/CERN/TIGRE/tree/axis-parallel) as I think I solved the 2D tomography issues here. In my tests, all works well, and you definitely do not need the `0.5` in the `geo.offDetector`,  it can be (shoudl be!) all zeroes. 

As you seem to be doing 2D tomography, I though of you as the best person to test this! 
(My various test worked well, but its always good to have someone else test).

genusn:
Ignored 0.001 corresponds to 0.001/SID radians. It's negligible, of course. :-)

AnderBiguri:
@genusn ah, yes, its its negligible for humans, my only worry is for this to cause some other numerical issue in some other edge-case.... But yeah, for now I can't really think of anything wrong happening with this micro-error that I introduce here....

PR #220, opened 2020-12-13
New faster way of computing matrix V (MATLAB)

yliu88au:
Here is a faster way to compute V by back-projection ones, it also allows larger problem to be solved, particularly for "OS" versions of the algorithms.


AnderBiguri:
Hi @yliu88au ,

I think you may have wanted to make a pull request from https://github.com/yliu88au/TIGRE/tree/yliu88au-patch-2 but instead you have made one from https://github.com/yliu88au/TIGRE/tree/master

This one has OS_AwASD_POCS, but not the compute V stuff. 

Ah, git is complicated to learn, no worries :)

yliu88au:
Could you "kill" this request? I need to harmonize my branch with mater first......

AnderBiguri:
Closed !

PR #221, opened 2020-12-13
Faster way computing V matrix by back-projection

yliu88au:
Faster way to compute V matrix by back-projection and fix a few bugs. Python version algorithms still have problems

### DIFF
--- a/Python/tigre/algorithms/iterative_recon_alg.py
+++ b/Python/tigre/algorithms/iterative_recon_alg.py
@@ -210,7 +210,7 @@ class IterativeReconAlg(object):
                 geox = copy.deepcopy(self.geo)
                 geox.angles = self.angleblocks[i]
                 # shrink the volume size to avoid zeros in backprojection
-                geox.sVoxel = geox.sVoxel * np.max(geox.sVoxel[1:] / np.linalg.norm(geox.sVoxel[1:]))
+                geox.sVoxel = geox.sVoxel * np.max(geox.sVoxel[1:] / np.linalg.norm(geox.sVoxel[1:])) * 0.9
                 geox.dVoxel = geox.sVoxel / geox.nVoxel
                 proj_one = np.ones((len(self.angleblocks[i]), geo.nDetector[0], 
                                     geo.nDetector[1]), dtype=np.float32)
--- a/Python/tigre/algorithms/iterative_recon_alg.py
+++ b/Python/tigre/algorithms/iterative_recon_alg.py
@@ -142,6 +142,10 @@ class IterativeReconAlg(object):
             'log_parameters',
             'angleblocks',
             'angle_index',
+            'alpha',
+            'alpha_red',
+            'rmax',
+            'maxl2err',
             'delta',
             'regularisation',
             'tviter',
@@ -180,18 +184,14 @@ class IterativeReconAlg(object):
         :return: None
         """
         geox = copy.deepcopy(self.geo)
-        geox.sVoxel[0:] = self.geo.DSD - self.geo.DSO
-        geox.sVoxel[2] = max(geox.sDetector[1], geox.sVoxel[2])
+        geox.sVoxel[1:] = geox.sVoxel[1:] * 1.11 # a bit larger to avoid zeros in projections
+        geox.sVoxel[0] = max(geox.sDetector[0],geox.sVoxel[0])
+        # geox.sVoxel[0:] = self.geo.DSD - self.geo.DSO
+        # geox.sVoxel[2] = max(geox.sDetector[1], geox.sVoxel[2])
         geox.nVoxel = np.array([2, 2, 2])
         geox.dVoxel = geox.sVoxel / geox.nVoxel
-        W = Ax(
-            np.ones(
-                geox.nVoxel,
-                dtype=np.float32),
-            geox,
-            self.angles,
-            "Siddon")
-        W[W <= min(self.geo.dVoxel / 4)] = np.inf
+        W = Ax(np.ones(geox.nVoxel, dtype=np.float32), geox, self.angles, "Siddon")
+#        W[W <= min(self.geo.dVoxel / 4)] = np.inf
         W = 1. / W
         setattr(self, 'W', W)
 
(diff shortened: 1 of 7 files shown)


Python/tigre/algorithms/iterative_recon_alg.py
```diff
@@ -352,7 +331,7 @@ def update_image(self, geo, angle, iteration):
         VERBOSE:
          for j in range(angleblocks):
              angle = np.array([alpha[j]], dtype=np.float32)
-             proj_err = proj[angle_index[j]] - Ax(res, geo, angle, 'Siddon')
+             proj_err = proj[angle_index[j]] - Ax(res, geo, angle, 'ray-voxel')
```
genusn, line 334: ```ray-voxel``` is supported but obsolete. Could you please revert this line?

Python/tigre/utilities/Ax.py
```diff
@@ -18,7 +18,7 @@ def Ax(img, geo, angles, projection_type="Siddon"):
     geox.cast_to_single()
     #geox.checknans()
 
-    if all(img.shape != geox.nVoxel):
+    if abs(img.shape - geox.nVoxel).max()>1e-8:
```
genusn, line 21: ```img.shape```  and ```geox.nVoxel``` are both supposed to be integers, so I don't think tolerance is necessary. If you are worried about the case where nVoxel is floating point value, how about checking like https://github.com/CERN/TIGRE/blob/41ffa73421512b49515e749a7d809d8d8c0af948/Python/tigre/utilities/Ax.py#L7 ? I think it is clearer to understand.

yliu88au: I think the problem I had is, after applied the conversion of geox to float32, geox.nVoxel becomes float32 as well, so it is not comparable to img.shape (int).... @genusn

Python/tigre/algorithms/iterative_recon_alg.py
yliu88au: Done

Python/tigre/utilities/Ax.py
AnderBiguri: Hum,  `nVoxel` cast to `float32` at this stage, weird design decision (from my side). Its then later converted to int again, as the CUDA code only allows for ints there 

https://github.com/CERN/TIGRE/blob/41ffa73421512b49515e749a7d809d8d8c0af948/Python/tigre/Source/_types.pxd#L5

In any case, @genusn is correct here. `float32` may nor be able to have infinite precision for decimal numbers, but any _integer_ up to `2^31` should be accurately represented, so the equality applies.

Python/tigre/algorithms/pocs_algorithms.py
```diff
@@ -162,8 +162,7 @@ def run_main_iter(self):
                 dtvg = dtvg * self.alpha_red
 
             self.beta *= self.beta_red
-            c = np.dot(dg_vec.reshape(-1,), dp_vec.reshape(-1,)) / \
-                max(dg * dp, 1e-6)
+            c = np.dot(dg_vec.ravel(), dp_vec.ravel()) / (dg * dp)
```
AnderBiguri, line 165: I need to think about this. They do the same thing in practice, but if I dont understand wrong, `ravel` _may_ return a copy, while `reshape(-1,0)` is ensured not to return a copy.  

As usual, if this is the case, we come to the speed vs memory efficiency situation. `ravel` will return a copy if the array is not contiguous in memory, meaning the `dot` would be slower. A copy will make `dot` faster. But then we need a couple of copies of the image, and this can be bad at large scale....

Let me think about this.

Python/tigre/utilities/Ax.py
yliu88au: @AnderBiguri That is just my quick "fix" to allow me to test other issue. I am sure there are better ways to do it. Feel free to change this.  I am still testing python version algorithms. SIRT seems not working, and for small problems (2D and 3D cases), POCS family seems OK. Problem occurs when we have large cases. However, MLEM works well in any case.

Python/tigre/utilities/im3Dnorm.py
```diff
@@ -34,6 +34,7 @@ def im3DNORM(img,normind,varargin=None):
     if type(normind) is int:
         return np.linalg.norm(img.ravel(),normind)
     if normind == 'TV':
-        raise Exception('TV not yet implemented')
-
+        gx,gy,gz = np.gradient(img.ravel())
+        g = np.sqrt(gx*gx + gy*gy + gz*gz)
```
AnderBiguri, line 38:
```suggestion
        g = np.sum(np.sqrt(gx*gx + gy*gy + gz*gz))
```

Python/tigre/utilities/im3Dnorm.py
```diff
@@ -34,6 +34,7 @@ def im3DNORM(img,normind,varargin=None):
     if type(normind) is int:
         return np.linalg.norm(img.ravel(),normind)
     if normind == 'TV':
-        raise Exception('TV not yet implemented')
-
+        gx,gy,gz = np.gradient(img.ravel())
```
AnderBiguri, line 37: This doesnt work for 2 reasons. 

1. if you `ravel` the array, then it doesn't have `gy` and `gz`, its 1D.
2. `np.gradient` uses central differences. A checkerboard pattern will have zero TV in this case. We need "the same" TV as its used in the source, i.e. backwards differences.

Python/tigre/utilities/Ax.py
AnderBiguri: Sure @yliu88au , I am happy to fix all the notes I am leaving in this code if you want. I still will leave them, otherwise I will forget :)

genusn: @yliu88au Thank you for explaining the reason. I didn't see geox.cast_to_single(). I propose moving this shape check block to the head of the function Ax.

Python/tigre/utilities/im3Dnorm.py
yliu88au: I see. Then maybe we should go back to np.diff, following MATLAB.

Python/tigre/utilities/Ax.py
yliu88au: @genusn I agree, maybe we should check the dimension of img at the beginning using geo.nVoxel, instead of geox.nVoxel (which is converted to float32).

Python/tigre/algorithms/iterative_recon_alg.py
```diff
@@ -180,18 +184,14 @@ def set_w(self):
...
-        W = Ax(
-            np.ones(
-                geox.nVoxel,
-                dtype=np.float32),
-            geox,
-            self.angles,
-            "Siddon")
-        W[W <= min(self.geo.dVoxel / 4)] = np.inf
+        W = Ax(np.ones(geox.nVoxel, dtype=np.float32), geox, self.angles, "Siddon")
+#        W[W <= min(self.geo.dVoxel / 4)] = np.inf
```
AnderBiguri, line 194: Are you sure this is not needed? I think it should stay.

AnderBiguri:
I have commited all suggestions that we made, and made a couple more of mine, but I can't really test the python code now. 
I am having trouble in my laptop with python 3.8.2 and import modules. Can anyone test this before I merge?

yliu88au:
@AnderBiguri I have tested a few samples on Python 3.8.5 and it seems running OK.

AnderBiguri:
@yliu88au thanks for letting me know. I suspect its my laptopn python+windows configuration that is not correct, rather than TIGRE... I may give it a go with conda. 

Have you tested the TV `im3Dnorm`?  the rest of the things are small changes, I am most afraid of that one, as I wrote it without even running it :)

Issue #223, opened 2020-12-15
POCS algorithms behave differently on MATLAB and python


AnderBiguri:
This has been reported. Investigate.

yliu88au:
@AnderBiguri I can see three issues here:
(1) FDK algorithm returns two different results from MATLAB and Python versions;
(2) 2D case specific: Remember in MATLAB version of minTV and AwminTV, we need to duplicate and pad 2D [m,n] array as [m,n,2] before we passing it to GPU to process. And then trim the returned result back to [m,n] array. In Python, 2D case will have [1,m,n] array, which is passed to GPU directly for processing. Question is, should we also duplicate and pad to [2,m,n] array in _minTV.pyx and _AwminTV, and then trim back?
(3) Since the GPU code for both version are same POCS_TV.cu and POCS_TV2.cu, my question is, will the returned result from GPU consistent when we feed it with [2,m,n] array or [m,n,2] array?

yliu88au:
Example code for 2D and 2D.

[example.zip](https://github.com/CERN/TIGRE/files/5713358/example.zip)

yliu88au:
@AnderBiguri I think I have fixed issues (2) and (3) in PR 228. Issue (1) still remains.

AnderBiguri:
Thanks for this comments and the PR!

As tyou say in (3), we might need to change a couple of things in the cuda source. It would be interesting to test with m==n first and compare both python and MATLAB to see if that is the same, and then change it. I suspect as you say (and I commented in the PR) that this has a negative influence. 

I opened #229 for FDK

PR #228, opened 2020-12-18
Fix some discrepancies btw MATLAB/Python

yliu88au:
Two type of changes: 
(1) Python passes the same matrix to GPU for POCS minimization as in MATLAB, makes the results closer to that of MATLAB version; 
(2) Changes to MATLAB version of POCS minimization for 2D cases, no need to duplicate before passing to GPU for processing.

### DIFF
--- a/Python/tigre/Source/_AwminTV.pyx
+++ b/Python/tigre/Source/_AwminTV.pyx
@@ -31,7 +31,7 @@ def cuda_raise_errors(error_code):
         raise ValueError('TIGRE:Call to aw_pocs_tv failed')
 
 def AwminTV(np.ndarray[np.float32_t, ndim=3] src,float alpha = 15.0,int maxiter = 100, float delta=-0.005):
-    src=np.transpose(src,(1,2,0)).copy()  # shift 1st dim to last, match MATLAB order
+#    src=np.transpose(src,(1,2,0)).copy()  # shift 1st dim to last, match MATLAB order
     cdef np.npy_intp size_img[3]
     size_img[0]= <np.npy_intp> src.shape[0]
     size_img[1]= <np.npy_intp> src.shape[1]
@@ -39,10 +39,10 @@ def AwminTV(np.ndarray[np.float32_t, ndim=3] src,float alpha = 15.0,int maxiter 
 
     cdef float* c_imgout = <float*> malloc(size_img[0] *size_img[1] *size_img[2]* sizeof(float))
 
-    cdef long imgsize[3]
-    imgsize[0] = <long> size_img[0]
+    cdef long imgsize[3] # shift 1st dim to last, match MATLAB order
+    imgsize[0] = <long> size_img[2]
     imgsize[1] = <long> size_img[1]
-    imgsize[2] = <long> size_img[2]
+    imgsize[2] = <long> size_img[0]
 
     cdef float* c_src = <float*> src.data
     cdef np.npy_intp c_maxiter = <np.npy_intp> maxiter
@@ -50,4 +50,5 @@ def AwminTV(np.ndarray[np.float32_t, ndim=3] src,float alpha = 15.0,int maxiter 
     imgout = np.PyArray_SimpleNewFromData(3, size_img, np.NPY_FLOAT32, c_imgout)
     PyArray_ENABLEFLAGS(imgout, np.NPY_OWNDATA)
 
-    return np.transpose(imgout,(2,0,1)).copy()
+    return imgout
+#    return np.transpose(imgout,(2,0,1)).copy()
--- a/Python/tigre/Source/_AwminTV.pyx
+++ b/Python/tigre/Source/_AwminTV.pyx
@@ -31,6 +31,7 @@ def cuda_raise_errors(error_code):
         raise ValueError('TIGRE:Call to aw_pocs_tv failed')
 
 def AwminTV(np.ndarray[np.float32_t, ndim=3] src,float alpha = 15.0,int maxiter = 100, float delta=-0.005):
+    src=np.transpose(src,(1,2,0)).copy()  # shift 1st dim to last, match MATLAB order
     cdef np.npy_intp size_img[3]
     size_img[0]= <np.npy_intp> src.shape[0]
     size_img[1]= <np.npy_intp> src.shape[1]
@@ -49,4 +50,4 @@ def AwminTV(np.ndarray[np.float32_t, ndim=3] src,float alpha = 15.0,int maxiter 
     imgout = np.PyArray_SimpleNewFromData(3, size_img, np.NPY_FLOAT32, c_imgout)
     PyArray_ENABLEFLAGS(imgout, np.NPY_OWNDATA)
 
-    return imgout
+    return np.transpose(imgout,(2,0,1)).copy()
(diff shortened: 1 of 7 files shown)


Python/tigre/Source/_AwminTV.pyx
```diff
@@ -31,6 +31,7 @@ def cuda_raise_errors(error_code):
         raise ValueError('TIGRE:Call to aw_pocs_tv failed')
 
 def AwminTV(np.ndarray[np.float32_t, ndim=3] src,float alpha = 15.0,int maxiter = 100, float delta=-0.005):
+    src=np.transpose(src,(1,2,0)).copy()  # shift 1st dim to last, match MATLAB order
```
AnderBiguri, line 34: I agree that this makes it the same as MATLAB, but I am suprised this makes a difference. 

In theory, the CUDA code here does not care about what is X,Y,Z axis, so my assumption would be that this line should do nothing, the output should be the same with and without it. 

If this is not the case, then I am a bit confused....

AnderBiguri:
When approved, this will fix #223    

(comment for automatic closure)

Python/tigre/Source/_AwminTV.pyx
genusn: @yliu88au Great work! Two questions:
1. Could you please try anisotropic case like
```
geo.nVoxel = np.array([1, 256, 512])
```
to see which of 
```
src=np.transpose(src,(1,2,0)).copy()
```
and 
```
src=np.transpose(src,(2,1,0)).copy()
```
is correct?

2. Did you try swapping like as follows
```
    cdef long imgsize[3]
    imgsize[0] = <long> size_img[2]
    imgsize[1] = <long> size_img[1]
    imgsize[2] = <long> size_img[0]
```
instead of ```transpose``` and ```copy```?

I'm sorry I cannot test by myself now.

AnderBiguri: I think @genusn second solution is what can make MATLAB and python do the same without the `np.transpose` call. 
Unfortunately I have not managed to get my python working yet so I can not test it yet! Its also holiday season where I am from so I will be a bit more sparse these days. But there is no hurry, we can fix this in few weeks otherwise.

yliu88au: This is is very interesting. I have tested @genusn second solution and it works for my particular 2D example. However, if I use the following swapping or without swapping, it won't work.


    imgsize[0] = <long> size_img[2]

    imgsize[1] = <long> size_img[0]

    imgsize[2] = <long> size_img[1]

The difference between with or without swapping is very significant. The POCS stop criteria "c" changes from
c=-0.21086 (without) to c=-0.9309 (with swapping). MATLAB version c=-0.96..., The remaining difference may be due to FDK initialization difference between Python and MATLAB version.

yliu88au: Further test using the same initial image (all zeros) shows that both MATLAB and Python versions yield very similar "c" values for the OS_AwASD_POCS algorithm (MATLAB c=-0.821. Python c=-0.831) for my 2D real data example. Happy with this result.

yliu88au: Continuing test shows that the following two types of working swapping in Python gave identical results for "square" case, ie., nVoxel=[h,n,n], where h can be 1 (2D case) or h>1 (3D case):
swapping 1: @genusn 2.
 
swapping 2:

    imgsize[0] = <long> size_img[1]
    imgsize[1] = <long> size_img[2]
    imgsize[2] = <long> size_img[0]

For non-squared case (for example, 128x256x512), the results are slightly different. It is a bit hard for me to judge which one is correct. However, it seems @genusn 's swapping is slightly better for both 2D and 3D cases.

AnderBiguri: I am almost certain that the correct one is:

```
imgsize[0] = <long> size_img[2]
imgsize[1] = <long> size_img[1]
imgsize[2] = <long> size_img[0]
```

Did this not work?

yliu88au: This does work, and it is the better one.

genusn: How about here? 
https://github.com/CERN/TIGRE/blob/41ffa73421512b49515e749a7d809d8d8c0af948/Python/tigre/Source/_tvdenoising.pyx#L53-L55

yliu88au: @genusn Good point, I will look at it and do some tests.

yliu88au: @genusn I have made the same change as in others to _tvdenoising.pyx:

    imgsize[0] = <long> size_img[2]
    imgsize[1] = <long> size_img[1]
    imgsize[2] = <long> size_img[0]

The test results seems make no difference here (still has small difference between Python and MATLAB, due to the different FDK init  images? #229 ). But for sake of consistency, we should make this change as well.

genusn:
As for the change 4629603 in this PR, I made a test as below:

1. Preparation on matlab:
```
%%
clear;
close all;

%% Prepare image
geo=defaultGeometry('nVoxel',[100;201;303]);                     
angles=linspace(0,2*pi,100);
head=headPhantom(geo.nVoxel);
noise_projections=addCTnoise(projections);
imgFDK=FDK(noise_projections,geo,angles);
%% Save image
save('matFDK.mat', 'imgFDK', '-v7');
%% TV denoise
imgdenoised=im3DDenoise(imgFDK,'TV',100,15);
%% Save denoised image
save('matFDK_denoised.mat', 'imgdenoised', '-v7');
```
2. Test on Python:
```
#%%
import scipy as sp
import scipy.io
import numpy as np
import tigre


#%%
pathFDK = "../MATLAB/matFDK.mat"
pathFDK_denoised = "../MATLAB/matFDK_denoised.mat"

obj_matFDK = scipy.io.loadmat(pathFDK)
obj_matFDK_denoised = scipy.io.loadmat(pathFDK_denoised)

print(type(obj_matFDK))
print(obj_matFDK.keys())
print(obj_matFDK['imgFDK'].shape)

imgFDK_mat = obj_matFDK['imgFDK'].transpose((2,1,0)).copy() # shape must be reorganized as (303, 201, 100)
imgFDK_denoised_mat = obj_matFDK_denoised['imgdenoised'].transpose((2,1,0)).copy()
print(imgFDK_mat.shape)
print(imgFDK_denoised_mat.shape)
# %%
imgFDK_denoised_py = tigre.utilities.im_3d_denoise.im3ddenoise(imgFDK_mat, 100, 15)
print(imgFDK_denoised_py.shape)

if imgFDK_denoised_py.shape == imgFDK_denoised_mat.shape:
    imgDiff = np.abs(imgFDK_denoised_py-imgFDK_denoised_mat)    # after
else:
    imgDiff = np.abs(imgFDK_denoised_py-imgFDK_denoised_mat.transpose((2,1,0)).copy()) # before

maxDiff = np.max(imgDiff)
minDiff = np.min(imgDiff)
print("diff is in [{}, {}]".format(minDiff, maxDiff))
#base(41ffa73) : diff is in [0.0, 0.7798717617988586]
#PR            : diff is in [0.0, 0.0]
```
The result was
 * The shape of the returned image from im3ddenoise used be in reverse order. Now it is the same as that of the input.
 * The difference between Python and MATLAB is zero.
  
Test env.
- Windows 10
- MATLAB: 2016a
    - MSVC2015   
- Python: anaconda 3.8
    - MSVC2019  
- GPU: GTX1060

AnderBiguri (approved):

Issue #197, opened 2020-11-17
MATLAB TIGRE failing in Quadro P5000


AnderBiguri:
Unfortunately this is second hand-information, so I can't really know exactly whats wrong, but someone with Quadro P5000 complains of "unspecified launch failure" errors, particularly on 2-GPU machines in MATLAB

this is interesting, as the error seems to be related to segfaults in the GPU code, but no one else seems to have (as far as I know) the same issue, so not sure what it may be. Sometimes this is caused by block/thread sizes etc. 

If anyone else has a P5000, can you test TIGRE there and report back to me, so I can have more info about this?

yliu88au:
I have had the similar problem on two-GPU (also Quadro P5000) before. After changed to newer version of CUDA and recompiled, the problem went away. @AnderBiguri

yliu88au:
@AnderBiguri Sorry, after further test, OS_AwASD_POCS (MATLAB) still crashes on the 2-P5000 machine. MLEM algorithm runs fine on the same computer (means Ax and Atb are good).

yliu88au:
Further investigation shows that algorithms crashes at line
f=minimizeTV(f0,dtvg,ng);  or f=minimizeAwTV(f0,dtvg,ng,delta); on a computer with 2 P5000 GPUs (MATLAB). Hope this information helps.

AnderBiguri:
@yliu88au yes that is actually very useful information. 

Its interesting because I have use those algorithms in 2 GPU machines well, but apparently either a particular configuration, or the P5000s are making this crash.

yliu88au:
With all new PR incorporated (PR221 and PR 228), **both** MATLAB and Python versions crash for 2D cases only and 3D case runs OK on this dual P5000 GPU computer, and with algorithms involving either POCS_TV.cu or POCS_TV2.cu. Same examples (2D and 3D cases of both versions) run OK on single GPU computers. @AnderBiguri

yliu88au:
@AnderBiguri Further tests show that, for problem with geo.nVoxel=[x,y,z], if z=1,2,3, POCS algorithms crash on both MATLAB and Python version on dual GPUs (P5000) computer. However, if z>=4, all runs OK. I am wondering if this is true for other dual GPU computers?

AnderBiguri:
@yliu88au it is likely true for most. I suspect a bug on the multi GPU code. That information helps a lot, its likely that the special case of small Z is not handled correctly when splitting the image among the gpus

[...]

PR #230, opened 2020-12-22
Fix crashing problem for Dual GPUs (#197)

yliu88au:
#197 A simple fix for POCS algorithms crash on dual GPU computer for both Python and MATLAB, when geo.nVoxel[2]<4. In these "thin" cases, POCS calculation are limited on single GPU.


AnderBiguri:
fix #197

Issue #219, opened 2020-12-10
Harmonize python/MATLAB demos


AnderBiguri:
Ideally we'd have the same examples in both. The python ones are particularly lacking.

yliu88au:
Good idea. So far I could not get python versions algorithms (particularly POCS) to agree with MATLAB version.

genusn:
As for demos, do you keep using jupyter notebooks or migrate to python scripts?

AnderBiguri:
@genusn still unsure. 

I really like jupyter notebooks when they are proper, as they can be rendered in Github, so one can have demos without even running them. But also they are a bit more of a paint to maintain and use in github. 

I though about making 2 set of "demos": demos and exercises. The first being basically the same as in MATLAB, using basic python scripts and the second just a limited amount of "shiny" visual examples, with tidier comments, plots etc using jupyter. As they currently are, there is no advantage of having the demos in jupyter, as they barely use the power of the notebooks (markdown, latex, plots etc). 


Of course, then I never get to do any of this, because its too many thins hehe.

AnderBiguri:
fair point. 

Thing is, MATLAB also has its own notebook system, that basically is a copy of what a jupyter notebook is. 
I think its likely better to have just all in standard MATLAB and python, then as you say, if at some point we have much nicer examples to show, we can either make a new repo  TIGRE-examples, or simply make a folder in Frontispiece with nice visual examples that work well in Github itself. 

This is what you meant right?

yliu88au:
@AnderBiguri I still could not get Python examples to agree with MATLAB for the POCS algorithms, even with the same setups I used. I have compared all CUDA sources (*.cu), the main difference between MATLAB and Python versions are lines:

"#if IS_FOR_MATLAB_TIGRE"
   "......"
"#else"
   "......"
"#endif"

My understanding of these lines are to deal with "row_major" and "column_major" difference in storing data in MATLAB and Python. However, I have found that there are 3 files: tvdenoising.cu, POCS_TV.cu, and POCSTV2.cu are identical for both MATLAB and Python. Do we still have the same row_major/column_major problem within these files?

AnderBiguri:
@yliu88au  do you have example code to reproduce this? Maybe we should open a different issue on the topic. I suspect its not the CUDA code, as the TV calculation is symmetric, thus axis-changing should not have an impact on the TV computation. Maybe something else is off...

AnderBiguri:
I have started working on this on #263

[...]

PR #232, opened 2020-12-25
Add plot_geometry.py

genusn:
* See #219

## Test
```
import numpy as np
import tigre

nVoxelZ = 128
nVoxelY = 256
nVoxelX = 512
nDetectorU = 400
nDetectorV = 300
offDetectorU = 0#500
offDetectorV = 0#500
offOriginX = 0#300
offOriginY = 0#100
offOriginZ = 0#100
geo = tigre.geometry(mode='cone',default=True)
geo.nVoxel = np.array([nVoxelZ,nVoxelY,nVoxelX])
geo.sVoxel = geo.nVoxel
geo.dVoxel = geo.sVoxel / geo.nVoxel
geo.nDetector = np.array([nDetectorV,nDetectorU])
geo.sDetector = geo.nDetector*geo.dDetector
geo.offDetector = np.array([offDetectorV, offDetectorU])
geo.offOrigin = np.array([offOriginZ, offOriginY, offOriginX])
print(geo)
angle = -np.pi/6
tigre.plot_geometry(geo, angle)
```

## Result

screenshot

Changing the parameters, I compared with the MATLAB version. I seemed OK, but I would appreciate it if you could reconfirm.

### DIFF
--- a/Python/tigre/__init__.py
+++ b/Python/tigre/__init__.py
@@ -5,5 +5,6 @@ from .utilities.Ax import Ax
 from .utilities.Atb import Atb
 from .utilities.plotproj import plotproj, plotProj, plotSinogram
 from .utilities.plotimg import plotimg, plotImg
+from .utilities.plot_geometry import plot_geometry
 from . import algorithms
 import tigre.demos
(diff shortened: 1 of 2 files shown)


AnderBiguri:
Fantastic, thanks! I'll double check this, but the MATLAB version is also not compete, so differences are not necessarily negative :)

PR #233, opened 2020-12-25
Curved 2D detector support

AnderBiguri:
This PR adds a function to "flatten" a curved detector. 

Missing

- [ ] test with real data
- [ ] Add demo

### DIFF
--- a/MATLAB/Utilities/flatten_detector.m
+++ b/MATLAB/Utilities/flatten_detector.m
@@ -11,7 +11,7 @@ pixel_arclength=arclength/size(proj,2);
 
 %% useful tools
 
-distance_arc_to_plane=@(alpha)((sqrt(1/(1-sin(alpha)^2))-1)*geo.DSD);
+distance_arc_to_plane=@(alpha)((sqrt(1/(1-sin(alpha).^2))-1)*geo.DSD);
 
 %% Geometric interesting values. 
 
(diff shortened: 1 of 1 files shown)


yliu88au:
I have done these conversions in a slightly different way. For my real world problem, the curved detector arrays are not exactly on a circle, rather, they are piece-wise linear on a curve (each piece-wise linear is corresponding to a detector module, many modules consist of the arc (blue in the graph)). I found the best way is to project all detector pixel on to a linear (wall, red in the graph) array, then interpolate the projection to a (larger) equal desistance linear array, as showing in the following graph. So it is basically interpolate the original sinogram to a slightly larger sinogram, before feeding into Tiger. It works well.

 
projection

AnderBiguri:
Hi @yliu88au, I think this is supposed to be doing the the same thing (that is what I intended to code anyway!) , but I haven't been able to test it yet because I don't have the data for it yet (it's coming).

I think my code assumes a continuous arc, and ignores that each detector pixels is flat, to allow for interpolation in between pixels. Unsure how much that will affect the results. 

I'd be more than happy to have a look at your way of doing it and try to incorporate it to Tigre, but you'd need to share some data with me so I can test it too (can be done privately).

AnderBiguri:
@yliu88au on top of the conversion to flat, I am also going to try to add data loaders for the machine that outputs it.

If your lab based system has a somehow standard way of outputting data i.e. Not just some raw data without any info, and you thunk it can be useful for others, we can also work on adding it to Tigre

yliu88au:
@AnderBiguri Unfortunately, the 2D data is from a third party and I cannot disclose it at all. My lab data are from flat panel detectors (3D), so it is irrelevant here.

AnderBiguri:
@yliu88au absolutely no problem :)

AnderBiguri:
This is on hold until I can get data to test it.

AnderBiguri:
https://gist.github.com/sjroe/1842e4fd44a6d4a8cb4eb163c4ce0d86

Issue #194, opened 2020-11-06
PEP-8 standards


AnderBiguri:
When the python code is somoehow final (not yet), we should have a look at enforce python style standards.

sco1:
Hi Ander, is this something you're open for collaboration on?

AnderBiguri:
Hi @sco1 ! 
Apologies, I do not know who you are (changed nickname since?).

But yes! absolutely, everything is open to collaboration :)

sco1:
I'm happy to work on the PR to add what's decided on. I'll put out some things to consider first, as there are some stylistic choices to make for a few in order to set up the desired linting errors to enforce.

#### Automatic Code Formatting with [black](https://github.com/psf/black)
This is an autoformatter to (mostly) PEP8 standards with one configuration option: line length. It may clash with some individual styles but it has the advantage of being consistent; basically run it and don't worry about differences in code style.

#### Code Linting with [flake8](https://flake8.pycqa.org/en/latest/)
In addition to the built-in errors, I'd like to suggest some additional plugins:

  * [flake8-docstrings](https://gitlab.com/pycqa/flake8-docstrings)
  * [flake8-import-order](https://github.com/PyCQA/flake8-import-order)
  * [flake8-bugbear](https://github.com/PyCQA/flake8-bugbear)
  * [pep8-naming](https://github.com/PyCQA/pep8-naming)

Of these, flake8-docstrings and pep8-naming probably have the biggest impact on style based on the selection of errors to enforce.

I'll also throw out [flake8-annotations](https://github.com/sco1/flake8-annotations) (caveat, I wrote this) if you're interested in adding type annotations to the codebase. Like [mypy](https://github.com/python/mypy), it also supports gradual addition of typing.

Here's the template I start with:

```
[flake8]
max-line-length=100
docstring-convention=all
import-order-style=pycharm
ignore=
    P102,B311,W503,E226,S311,
    # Missing Docstrings
    D100,D104,D105,D107,
    # Docstring Whitespace
    D203,D212,D214,D215,
    # Docstring Quotes
    D301,D302,
    # Docstring Content
    D400,D401,D402,D404,D405,D406,D407,D408,D409,D410,D411,D412,D413,D414,D416,D417,
    # Type Annotations
    ANN002,ANN003,ANN101,ANN102,ANN204,ANN206,
    # pep8-naming
    N802,N806,N815,
exclude=
    __pycache__,.cache,
    venv,.venv,
```

[...]

PR #237, opened 2020-12-31
Fix typo in docstring

genusn:
- lambdared => lmbda_red, lambda => lmbda
   - For your reference:
https://github.com/CERN/TIGRE/blob/c5d0050f5a87d5979524bbab9639d092266ea579/Python/tigre/algorithms/iterative_recon_alg.py#L127-L134

- I think that it would be nice to add a note that lmbda is not a typo.
   - or change the option name to step_size , for example.
   - or other...
   - but this is another issue.


AnderBiguri:
lambda reduction is not a step size, lambda itself is. Reduction is by how much it is slowed down. Maybe lamdba is not a very ovbious name I guess, but its generally called that in maths papers. What do you think?

genusn:
A happy new year.

Thank you for your comments and merging my PRs.

Hmm.... The first book that I read was Kak and Slaney's book (http://www.slaney.org/pct/pct-toc.html, p.284), and it uses 'relaxation parameter' $\alpha$. It cited Herman's pager, which seems to use $\lambda$.
Wiki (https://en.wikipedia.org/wiki/Gradient_descent) uses $\gamma$ for 'step size'.
Furthermore, in python, ```lmbda```, ```lmbda_red``` and ```tvlambda``` are correct, not ```lambda``` , ```lambda_red``` and ```tvlmbda``` respectively.

I think it is Okay to use alpha, beta, gamma, A, b, x, etc. inside the toolbox, but I don't think it's user-friendly to expose them.
However,  ```lambda``` is used both in MATLAB and Python from the beginning. I think it is difficult to change its name, but additional explanation such as 

```lmbda``` is the relaxation parameter if you use the default ```dataminimizing```, which is GD. Note that the parameter name is lmbda, not lambda because the word ```lambda``` is reserved by Python :-P

, for example of course, might be a possible solution.

AnderBiguri:
Happy new year to you too!

Indeed I did not  think in the future when assigning the name `lambda` in the MATLAB version hehe. 

I agree too, that many of the parameters have terrible names. I still don't know how to tackle the enormous list of TV parameters with random greek letters.  I was thinking on making some aliases, but I think it makes the code harder to maintain and uglier, so maybe we need to accept that the names are as they are. As you say, likely the best idea is to just add good explanations of what they are with other words, such as "step size" etc. 

This is also where the python demos are needed. I think that the MATLAB ones do a relatively good job in explaining these things in general.

genusn:
Agree. #219 and #194 seem to be related to this discussion.

PR #243, opened 2021-01-28
Compute 30 build

gfardell:
Added compiler macro for CUDA 9.2 for `cudaDeviceGetAttribute()` otherwise `isHostRegisterSupported = 0`

` __shfl_down_sync()`  needs CUDA 9.0, and 9.0 no longer supports compute capability 3.0 so I changed the compiler macro to catch this (was a check for compute capability 3.0)

Removed deprecation of compute 3.0 from setup.py

### DIFF
--- a/MATLAB/Source/POCS_TV.cu
+++ b/MATLAB/Source/POCS_TV.cu
@@ -183,7 +183,7 @@ do { \
         __syncthreads();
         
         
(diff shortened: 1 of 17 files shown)


AnderBiguri:
Fantastic! let me do CI (i.e does it run in my machine?) and merge then.

PR #250, opened 2021-02-01
Make MATLAB compilation CUDA-agnostic 

AnderBiguri:
Compile.m will now 

- Find your CUDA path as setup.py does
- Read your CUDA version
- Modify xml files to set the correct compute capabilities for your CUDA version


AnderBiguri:
tested on Ubuntu 18.04 and Windows 10.

Issue #244, opened 2021-01-28
Unify source


AnderBiguri:
We have 2 folders with source, and it seems that we could work something out to just have a single folder and build python or matlab against it, halfing our job when improving it, and minimizing future bugs.

genusn:
Good idea.

Before going further, I think the files to define mex functions such as minTV.cpp, tvDenoise.cpp and projections.cpp should be moved to MATLAB/Utilities or somewhere under MATLAB folder.

AnderBiguri:
My suggestion:

Have the main repo have an extra folder `/Source` or `/CUDA`. I think this makes more sense than to have python build against `MATLAB/Source` (or vice versa)

Then, I see a couple of options:

1.  Then, leave the `cpp` files in `/MATLAB/Source`.
2.  Completely delete the `/MATLAB/source` folder and have a `/MATLAB/Utilities/Source` folder with the `cpp` files

Any other proposal that we can consider?

genusn:
Almost agree. Let me comment two points:
1. Any .m file seems to be a source file to me, so ```CUDA``` would be better than ```Source```. Or if you think the .cu files are the core of this project or going to add other GPGPU technology such as OpenCL, ```Core```, ```Common```, ```Common/CUDA``` or ```Core/CUDA``` might be possible.
2. Under ```MATLAB/Utilities``` folder, there are ```IO``` and ```GPU``` folders that have their mex files inside. So, the destination of mex cpp files can be ```/MATLAB/Utilities``` instead of ```/MATLAB/Utilities/Source```.

AnderBiguri:
I agree. 

We have Python/MATLAB folders, the other one should be CUDA. If TIGRE ends up using some other GPU tech, then we can make another folder. 

Similarly, I agree with the name source there. Maybe something like `MATLAB/Utilities/mex_interface` or `./cuda_interface`

genusn:
```Python\tigre\Source``` and ```MATLAB\Source``` are compared.

Only in Python  => Python/tigre/utilities/cuda_interface
 * ```__init__.py```
 * ```_Atb.pyx```
 * ```_AwminTV.pyx```
 * ```_Ax.pyx```
 * ```_gpuUtils.pxd```
 * ```_gpuUtils.pyx```
 * ```_minTV.pyx```
 * ```_tvdenoising.pyx```
 * ```_types.pxd```
 * ```errors.hpp```
 * ```projection.cpp```
 * ```projection.hpp```

[...]

PR #253, opened 2021-02-07
Unify source -- a solution for #244

genusn:
PR for #244

## What's changed...

1. Checked the difference between two ```Source``` folders.
    * Three categories
        1. CUDA files. (.cu, .hpp, GpuIds.cpp)
             * No difference.
        2. Used only by Python
             * Only in Python/Source folder
             * Cython files: *.pyd, *.pyx
             * CUDA files: projection(.hpp,|.cpp)
        3. Used only by Matlab (other .cpp files)
             *  MEX functions are defined.
2. Reorganized the folder structure:
   * CUDA files
      * Moved from MATLAB/Source/ to Common/CUDA/
      * Removed from Python/Source/ after moving projection(.hpp,|.cpp)
   * Python files
      * Moved .pyd and .pyx files to Python/utilities/cuda_interface/
      * Moved projection(.hpp,|.cpp) to Common/CUDA/
   * Matlab files
      * Moved .cpp files to MATLAB/Utilities/cuda_interface/
3. Modified to compile successfully
   * Compile.m and Setup.py
   * Relative path of #include
   * Relative path of python modules.

## What's tested

1. Matlab
   * Run d04, d06, d07, d08, d09, d20.
      * Finished without error.

|**Windows**| 10  64bit |
|**MATLAB**|  2016a |
|**CUDA**| 11.1|
|**Visual Studio**| 2015|
|**GPU**| 2xRTX2080, 1xGTX1070 |

2. Python
   * Run example.py
      * Finished without error.

|**Windows**| 10  64bit |
|**Python(anaconda)**|  3.8.5 |
|**CUDA**| 11.1|
|**Visual Studio**| 2019 |
|**GPU**| GTX1060 |


AnderBiguri:
Thanks! I'll slowly start testing. 

Currently in linux/MATLAB, compilation errors:

```
 mex -largeArrayDims ./Utilities/GPU/getGpuName_mex.cpp ../Common/CUDA/gpuUtils.cu -outdir ./Mex_files/linux64
Building with 'nvcc'.
Error using mex
/home/anderbiguri/Documents/MATLAB/TIGRErepo/TIGRErepo/MATLAB/./Utilities/GPU/getGpuName_mex.cpp:2:10: fatal error:
../../../Common/Cuda/gpuUtils.hpp: No such file or directory
 #include "../../../Common/Cuda/gpuUtils.hpp"

          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

Likely an easy fix, I'll have a look later.

AnderBiguri:
I am not a C++ expert, but it seems that `include` with `../../something.hpp` is frowned upon in the C++ community and that instead we should use full paths `./TIGRE/MATLAB/Common/something.hpp` and then include the header location in the compilation (I think with `-I`, but in MATLAB I am sure it can also be added to the `xml` file). 

What do you think? 

Still unsure why it fails in linux.

genusn:
As for ```include```, I'll try.

As for linux issue, me too. I could not find any difference between win64 and linux64 in ```Compile.m```, but I cannot test because I have no access to Linux machine with GPUs. Is there any restriction of gcc around include, such as to prevent including files outside the workspace? I've found that ".." may cause C4464 in VC.

AnderBiguri:
I think its quite likely that the `..` is the issue, but I have not found anything that suggest me its not supported. The most I found is some comment in an stackoverflow post that discussed why `../..` is not recommended that suggested its also not cross-platform, but nothing more was mentioned. 

I was hoping that if we follow the suggestions of the `include` maybe the linux issue dissapears...

genusn:
@AnderBiguri I am adding ``` -I&quot;..\Common&quot;```  at ```config/vars@INCLUDE``` in ```mex_CUDA_win64.xml```. It seems to compile successfully with the following modification to Ax_mex.cpp as
```
#include <string.h>
#include <tmwtypes.h>
#include <mex.h>
#include <matrix.h>
#include <CUDA/ray_interpolated_projection.hpp>
#include <CUDA/ray_interpolated_projection_parallel.hpp>
#include <CUDA/Siddon_projection.hpp>
#include <CUDA/Siddon_projection_parallel.hpp>
#include <CUDA/GpuIds.hpp>
...
```
getGpuName_mex.cpp as
```
#include <mex.h>
#include <Cuda/gpuUtils.hpp>
```
etc.
Is this acceptable?

AnderBiguri:
Yes, sounds good!

genusn:
Ah, 
> #include "../../../Common/Cuda/gpuUtils.hpp"

I think ....CUDA is correct. Linux is case-sensitive, isn't it?

AnderBiguri:
Tested that, you are correct. That was the issue. Changing it to CUDA works in the MATLAB version.

AnderBiguri:
@genusn resolved one conflict for you. 

This seems good, let me do more thorough tests, but I don't see any issue. 
do you mind also adding the info of what happened to the recently introduced change-log? Feel free to add your real name/username to contributors and/or the changelog too.

genusn:
Thank you. I should have see your comment a few minutes ago...

changelog.md
```diff
@@ -6,7 +6,7 @@ Please, add here all changes added since previous version of TIGRE (v2.1)
 ## Features
 
 - You can now select which GPUs to run the code on, either by name, or GPUid. By @genusn
-- Improved compilation, now there is no need to edit setup.py/xml files for your GPU/CUDA SDK
+- Improved Python compilation, now there is no need to edit setup.py for your GPU/CUDA SDK
```
AnderBiguri, line 9: This seems to be rewriting backwards the changelog

MATLAB/Compile.m
```diff
@@ -28,212 +28,93 @@
...
-        mex  ./Utilities/GPU/getGpuName_mex.cpp ./Source/gpuUtils.cu -outdir ./Mex_files/linux32
-        mex  ./Utilities/GPU/getGpuCount_mex.cpp ./Source/gpuUtils.cu -outdir ./Mex_files/linux32
-    end
-end
-
-disp('')
-disp('Compilation complete')
-
-
-function [cuda_path, cuda_ver]=locate_cuda()
```
AnderBiguri, line 122: This is being removed!

AnderBiguri (changes requested):
Seems that for some reason there is been a mistake (maybe my fault in that master merge) and some changes of master are being removed in this PR. 

We should fix that.

MATLAB/mex_CUDA_glnxa64.xml
```diff
@@ -1,73 +1,74 @@
...
-    <vars
-          CMDLINE1="$CXX -c $DEFINES $INCLUDE $NVCCFLAGS $CXXFLAGS $OPTIM $SRC -o $OBJ"
-          CMDLINE2="$LDXX $LDFLAGS $LDTYPE $LINKOPTIM $LINKEXPORT $OBJS $CXXLIBS $LINKLIBS -o $EXE"
-          CMDLINE3="rm -f $OBJ"
-          
-          CXX="$NVCC"
-          DEFINES="--compiler-options=-D_GNU_SOURCE,$MATLABMEX"
-          MATLABMEX="-DMATLAB_MEX_FILE"
-          NVCCFLAGS="-gencode=arch=compute_35,code=sm_35  -gencode=arch=compute_50,code=sm_50   -gencode=arch=compute_60,code=sm_60   -gencode=arch=compute_61,code=sm_61 -gencode=arch=compute_70,code=sm_70 -gencode=arch=compute_75,code=sm_75  --default-stream per-thread  $NVCC_FLAGS"
-          CXXFLAGS="--compiler-options=-ansi,-fexceptions,-fPIC,-fno-omit-frame-pointer,-pthread, -Dnullptr=0 "
```
AnderBiguri, line 35: This one is still being replaced, which should not (`-Dnullptr=0`)

Python/setup.py
```diff
@@ -39,7 +39,7 @@ def get_cuda_version(cuda_home):
                 version_str = f.readline().replace('\n', '').replace('\r', '')
                 return version_str.split(" ")[2][:4]
         else:
-            version_str = subprocess.check_output([os.path.join(cuda_home,"bin","nvcc"),"--version"])
+            version_str = subprocess.check_output([os.path.join(cuda_home,"bin","nvcc.exe"),"--version"])
```
AnderBiguri, line 42: same, this is being reverted to previous state.  should be without exe

changelog.md
genusn: I didn't commit this file at the merger bb72dbe. Sorry. I should have done it.

MATLAB/mex_CUDA_glnxa64.xml
genusn: Fixed in defa305

Python/setup.py
genusn: Fixed in defa305

genusn:
Thank you for accepting my PR and sorry for merge failure and wasting your time.

AnderBiguri:
@genusn absolutely no worries! These things happens, you tend to do very nice PRs and its somehow likely it was my fault hehe. 

As usual, thanks a lot for this PR, it makes the repository considerably better.

Issue #284, opened 2021-04-16
improve timing printing in python


AnderBiguri:
reduce decimals, and dont just print (s) but a human readable time in H:M:S:ms

Ayushsunny:
Hello Sir @AnderBiguri, can you guide me on this? I would like to work on this issue

AnderBiguri:
Hi @Ayushsunny , absolutely! 

This one is quite easy:

Here: 

https://github.com/CERN/TIGRE/blob/master/Python/tigre/algorithms/art_family_algorithms.py#L103

and maybe a couple of other places, we use `toc-tic` to print the time, in seconds. Ideally, we just want some function that takes the time in seconds, and converts it to HH:mm:ss:ms  kind of format, so its easier to read. We want it to print something like:

"Esitmated time until completetion :  XXh XXm XXs XXXXms. "

Perhaps we can simply just remove h,m,s whenever they are 0.

Ayushsunny:
So I need to make this function right ?

```python
import time
time.strftime('%H:%M:%S', time.gmtime(12345))
'03:25:45'
```
but I didn't understand that 
`The best way is probably to make a new .py file within Utilities folder, that takes time in s and returns a string.
MATLAB has already a function (you can test it in MATLAB or Octave if you don;t have a MATLAB function):`

AnderBiguri:
@Ayushsunny ah, if there is an inbuilt function for it, then that is good enough!
In MATLAB there is not, so I was suggesting writting it up, but if `time.strftime` is what we want, the lets just use that! no need to make a new .py file :)

Ayushsunny:
> @Ayushsunny ah, if there is an inbuilt function for it, then that is good enough!
> In MATLAB there is not, so I was suggesting writting it up, but if `time.strftime` is what we want, the lets just use that! no need to make a new .py file :)

so this is the tic-toc function https://github.com/CERN/TIGRE/blob/bde3d952bbc63586928dd5f29e4fa1be30aed573/Python/tigre/algorithms/art_family_algorithms.py#L83-L104

we are going to change sir @AnderBiguri ?

AnderBiguri:
Specifically this part:

```
print( 
                         "Esitmated time until completetion (s): " 
                         + str((self.niter - 1) * (tic - toc)) 
                     ) 
```

I suggest running demo 4 of the python code, that will use this printing, and you will see it prints in `s` but we want a more human readable format.

AnderBiguri:
[...]

PR #290, opened 2021-05-04
added human readable time in HH:MM:SS:MS

Ayushsunny:
improved timing printing in python in HH:MM:SS:MS 
issue code - #284

here is the screenshot of output 

Screenshot_63


Ayushsunny:
Hello @AnderBiguri , could you review this PR please?

PR #299, opened 2021-05-19
Bruker data loader 

AnderBiguri:
Data Loader for Bruker Skyscan line of micro-CT machines


AnderBiguri:
Currently the code "works", but there are still geometric mistakes in the input, that I don't seem to be able to know how to solve. There are small per-projection shifts that are not obviously documented in the files and this cause halo-like errors.

AnderBiguri:
This works for datasets from 1 user, but not from dataset from a second user, which we think may be because of the data, and not because of the reader/

AnderBiguri:
MATLAB data loader still has no support for multiple datasets. It does for the geometry, but not the tiffs.

PR #333, opened 2021-09-27
Data Loaders for YXLON scans

AnderBiguri:
This data loader is verified by YXLON to work with their data, for most use cases ( helical and other non-standard not supported). 

However, I have no way to verify this yet, so I can not refactor it to TIGRE format or add a MATLAB version. 

If someone has a YXLON dataset, I would please ask you to get in contact with me so we can finish up the PR.


RyanhilCT:
Hi Ander,

I have a YXLON dataset available. I'm investigating iterative methods for application on real world large multi-material test objects. So having a MATLAB version data loader for YXLON would be useful. 

Let me know how you want the YXLON dataset.

AnderBiguri:
Hi @RyanhilCT

Thanks! Absolutely, it would be very useful, so I can quickly script something that may work. I suspect they are not too hard to read. 
If you could send me a private email tho arrange this it would be great!

AnderBiguri:
I still need more YXLON data loaders.


This PR contains a Loader in MATLAB and a Loader in Python. However:

- [ ] Python Loader does not fit TIGRE python data loader standard
- [ ] Python Loader (from YXLON staff) is radically different to MATLAB Loader
- [ ] MATLAB Loader (tested in YXLON data) is radically different to Python Loader

I need more datasets, as I can only access now a dataset that conforms with the MATLAB one. However, knowing that the python one was written by YXLON staff, I know its quite correct.  I assume its just an issue of different versions, but I would need more data to verify this before merging.

Issue #280, opened 2021-04-15
staticDetectorGeo() is missing to allow for DBT in python


AnderBiguri:

yliu88au:
Will this function work?

[staticDetectorGeo.zip](https://github.com/CERN/TIGRE/files/7667433/staticDetectorGeo.zip)

AnderBiguri:
@yliu88au  yes it should. I have no way of validating it, have you tried it with DBT data?

yliu88au:
I have only tested it on a simulated case of tomosynthesis (construct an image, then projecting and reconstruction), it all makes sense. But it is still not the real data.

AnderBiguri:
@yliu88au  if you simulated it outside TIGRE, then that is good enough! want to make a PR?

Issue #272, opened 2021-04-14
SART_TV not available in python


AnderBiguri:
requires #271

PR #352, opened 2021-12-10
Python algorithms tidy up

yliu88au:
Tidy up algorithms for Python:
(1) SART_TV added (#272)
(2) PCSD algorithm family added
(3) FISTA algorithm extended to allow even faster convergence (FISTA_mod)
(4) staticDetectorGeo added (#280)

The following total 18 algorithms for Python tested:
'FDK','SART','OSSART','SIRT','SART_TV','OSSART_TV','CGLS','FISTA',              'FISTA_mod','ASD_POCS','OS_ASD_POCS','AwASD_POCS','OS_AwASD_POCS',                 'PCSD','OS_PCSD','AwPCSD','OS_AwPCSD','MLEM'


AnderBiguri:
Before I do a review, there is a massive random number generator file, is it there by mistake?

yliu88au:
I don't know where it come from, just deleted that.

yliu88au:
Now FISTA and FISTA_mod is the same, but with different parameters. Also, I think, POCS family, including PCSD family, should be able to run as "OS" versions simply by specify blocksize = bsz, when bsz >1 when calling.  All "OS" version is to allow blocksize = 20 as default.

AnderBiguri:
Yes, the OS part is always something I wonder how to tackle. Historically and in some research circles, OS algorithms are treated very differently than per-projection algorithms. There is a point mathematically, as convergence properties are indeed different, however programatically, its basically just chosing a block size, and that will give you e.g. SART, OS-SART and SIRT. 

So I am not sure what to do API-wise. It makes sense to have SART be able to do block sizes, but at the same time, SART is defined as to be 1 by 1. It gets muddier in ASD-POCS, PCSD etc, as the math of this does not require them to be projection by projection, but the articles that published them do published them as such, and many users are using them to compare with the articles. So, theoretically ASD-POCS with OS-SART update is still ASD-POCS (instead of OS-ASD-POCS, which I made up).

I think the best way, for consistency with publications and so people don't make mistakes is to to have PCSD be only bsz=1, and then have a OS_PCSD version, similar with the other algorithms

yliu88au:
Yeah, that makes sense. Programmatically, it will be easy to define all algorithms in OS version as the starting class, then non OS version will simply inherit it with the blcoksize limited to 1. Anyway, it is a matter of choice.

AnderBiguri:
Agree, as long as we give them separate names, and don't allow  users to call non OS as OS, its all good.

AnderBiguri (approved):

PR #358, opened 2022-03-07
Fix offset weighting bug and suppress indexing warnings

twhitbread:
- _MATLAB/Algorithms/FDK.m_: If Wang weights are used, we need to use the new offset corresponding to the zero-padded detector.
- Suppressed command line warnings due to non-integer indexing (I think Matlab automatically does `round` in these cases anyway).


AnderBiguri:
Let me test this to double check and I'll merge it.

PR #360, opened 2022-03-17
Nikon Data loader patch

rjkoch:
Fixed issues with Nikon data loader. number of pixels and voxels should be integers, not float. Index, for loading images, should start at 0, not 1. Used enumerate rather than iterating the index in the loop.

### DIFF
--- a/Python/tigre/utilities/io/NikonDataLoader.py
+++ b/Python/tigre/utilities/io/NikonDataLoader.py
@@ -60,7 +60,7 @@ def readXtekctGeometry(filepath):
 
     ## Detector information
     # Number of pixel in the detector
-    geometry.nDetector = numpy.array((float(cfg["DetectorPixelsX"]), float(cfg["DetectorPixelsY"])))
+    geometry.nDetector = numpy.array((int(cfg["DetectorPixelsX"]), int(cfg["DetectorPixelsY"])))
     # Size of pixels in the detector
     geometry.dDetector = numpy.array(
         (float(cfg["DetectorPixelSizeX"]), float(cfg["DetectorPixelSizeY"]))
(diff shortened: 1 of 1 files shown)


Python/tigre/utilities/io/NikonDataLoader.py
```diff
@@ -157,9 +158,11 @@ def loadNikonProjections(folder, geometry, angles, **kwargs):
     image = numpy.asarray(image).astype(numpy.float32)
     projections = numpy.zeros([len(indices), image.shape[0], image.shape[1]], dtype=numpy.single)
     projections[0, :, :] = -numpy.log(image / float(geometry.whitelevel))
-    index = 1
+    #Python uses zero-based indexing
```
AnderBiguri, line 161: Yes, but we just read the "zeroeth" image in the previous line! so we can skip reading it!

AnderBiguri (changes requested):

Python/tigre/utilities/io/NikonDataLoader.py
rjkoch: Sure, that's fine, but in the previous version, `index `eventually ends up one larger than the size of `projections`

You could skip the first projections with
`for index, i in enumerate(tqdm(indices[1:]),1):`

AnderBiguri: Did it? Weird, I thought I had this well tested. 
Sure the change you suggest make sense. Mind making a commit? then I'll merge the PR.

rjkoch:
I'll get back into this today after my day job. But yes, I'll make another
PR

AnderBiguri:
@rjkoch  Thanks! No need to make another PR, just make a commit to rjkoch:master, it should appear in this PR too.

rjkoch:
Got it, sorry. New to git stuff. I made these changes to my local version,
it was the only way I could get it to run on a test dataset.

I'll see if I can pull a test data set for use here.

AnderBiguri:
@rjkoch No worries! All this git stuff is not trivial, particularly when you are starting :) 
I have some Nikon datasets myself, so once you make the change we discussed I'll also test it on my side, make sure all runs well

rjkoch:
made the suggested changes.

rjkoch:
Just checked the unchanged loader on a dataset. With:
`proj, geo, angles  = tigreio.NikonDataLoader(rel_dir.__str__(),sampling='equidistant',num_angles=150)`

I get:
`IndexError: index 154 is out of bounds for axis 0 with size 154`

This is because `projections` has shape `(n,I,J)` where `n` is the number of projections to load. but `index`, used to slice `projections`, ends up at `n`  and we can't slice an array of size `n` using `n` but rather at most `n-1`

Issue #375, opened 2022-06-16
d01_CreateGeometry.py raises an AttributeError: 'Arrow3D' object has no attribute 'do_3d_projection'


LEECHAN96:
## Expected Behavior
Create geometry and show it

## Actual Behavior
Traceback (most recent call last):
  File "C:\ProgramData\Anaconda3\envs\torch\lib\site-packages\matplotlib\backends\backend_qt.py", line 477, in _draw_idle
    self.draw()
  File "C:\ProgramData\Anaconda3\envs\torch\lib\site-packages\matplotlib\backends\backend_agg.py", line 436, in draw
    self.figure.draw(self.renderer)
  File "C:\ProgramData\Anaconda3\envs\torch\lib\site-packages\matplotlib\artist.py", line 73, in draw_wrapper
    result = draw(artist, renderer, *args, **kwargs)
  File "C:\ProgramData\Anaconda3\envs\torch\lib\site-packages\matplotlib\artist.py", line 50, in draw_wrapper
    return draw(artist, renderer)
  File "C:\ProgramData\Anaconda3\envs\torch\lib\site-packages\matplotlib\figure.py", line 2837, in draw
    mimage._draw_list_compositing_images(
  File "C:\ProgramData\Anaconda3\envs\torch\lib\site-packages\matplotlib\image.py", line 132, in _draw_list_compositing_images
    a.draw(renderer)
  File "C:\ProgramData\Anaconda3\envs\torch\lib\site-packages\matplotlib\artist.py", line 50, in draw_wrapper
    return draw(artist, renderer)
  File "C:\ProgramData\Anaconda3\envs\torch\lib\site-packages\mpl_toolkits\mplot3d\axes3d.py", line 451, in draw
    for artist in sorted(collections_and_patches,
  File "C:\ProgramData\Anaconda3\envs\torch\lib\site-packages\mpl_toolkits\mplot3d\axes3d.py", line 426, in do_3d_projection
    signature = inspect.signature(artist.do_3d_projection)
AttributeError: 'Arrow3D' object has no attribute 'do_3d_projection'

## Code to reproduce the problem (If applicable)
(the demo1 code)
```

```

## Specifications 
  - Python version: 3.8.8 64bit, in a conda environment
  - MATLAB version: matplotlib 3.5.2
  - OS: Win 10
  - CUDA version: 10.2

AnderBiguri:
[...]

PR #377, opened 2022-06-23
do_3d_projection method in Arrow3D

harshitAgr:
Added do_3d_projection method in Arrow3D inside plot_geometry.py. 
source:  https://github.com/matplotlib/matplotlib/issues/21688


AnderBiguri:
Thanks! Fix #375

PR #385, opened 2022-07-20
Improve python installation guide

pjueon:
- add `setuptools` to Python/requirements.txt 
- add `pip install -r requirements.txt` to the compilation guide.

There are packages that are required to compile TIGRE (ex> `Cython`, `setuptools`).
You should install them before you execute `setup.py` or you'll get the import error.


AnderBiguri:
Fantastic! I was hoping to get this done soon, but you saved me a bit of time, thanks a lot!

PR #388, opened 2022-07-21
update .gitignore

pjueon:
I found that `Python/tigre/utilities/cuda_interface/_randomNumberGenerator.cpp` was not included in `.gitignore` .
So I checked the `.gitignore` file and updated it.

Now, all the .cpp files in `cuda_interface`, including `_randomNumberGenerator.cpp`, are added to `.gitignore`.
So you don't have to edit `.gitignore` any more even if any other new .cpp files are added in the future.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,12 +17,7 @@ Python/purge.sh
 Python/.idea/
 Python/devtest.py
 Python/dist/
-Python/tigre/utilities/cuda_interface/_Atb.cpp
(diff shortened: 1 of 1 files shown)


AnderBiguri:
Fantastic! 
I know these are small changes, but they do really make a difference to keep TIGRE tidy, so I really appreciate the effort!

pjueon:
You are welcome :) 
And thank you for making this wonderful project!

PR #389, opened 2022-07-22
remove unnecessary code

pjueon:
In the demo 16 of python, the offsets were already set.


AnderBiguri:
Hi! 
You are correct, but those lines are there more to showcase that you can do it, rather than to have functioning code. 
Maybe the change required is to clarify that in the comments, something like 

`# Offsets (Optional, the default values are zero, thus there is no need to define them if you have none). `

pjueon:
You mean I removed all the lines that set offsets and am using the default values of the offsets?
No, what I was trying to say was there were exactly same lines right above.

AnderBiguri:
Ah, I checked the file wrong, Accidentaly oppened your modified version rather than TIGRE, you are absolutely right.

PR #396, opened 2022-08-18
Pull request for Krylov subspace algorithms

malena-sabate:


MATLAB/Algorithms/LSQR.m
```diff
@@ -0,0 +1,273 @@
...
+u=proj-Ax(x,geo,angles,'Siddon','gpuids',gpuids);
+normr = norm(u(:),2);
+u = u/normr;
+
+beta = normr;
+phibar = beta;
+
+% msl: I have not checked 'redundancy_weights', do they have to do with 
+% mismatched transpose? 
+if redundancy_weights
```
AnderBiguri, line 78: No, we should remove them. They are weigths that minimize overlap of data when there is offset of detector. They are quite heuristic, so I think they will screw up our nice Krylov spaces.

MATLAB/Algorithms/LSQR.m
```diff
@@ -0,0 +1,273 @@
...
+    x0 = x;
+    if (ii==1 && verbose);tic;end
+    
+    % (3)(a)
+    u = Ax(v,geo,angles,'Siddon','gpuids',gpuids) - alpha*u;
+    beta = norm(u(:),2);
+    u = u / beta;
+    
+    % (3)(b)
+    if redundancy_weights % msl: Is this right here?
```
AnderBiguri, line 106: same stuff here.

MATLAB/Algorithms/LSQR.m
```diff
@@ -0,0 +1,273 @@
...
+    end
+    alpha = norm(v(:),2);
+    v = v / alpha;    
+
+    % (4)(a-g)
+    rho = sqrt(rhobar^2 + beta^2);
+    c = rhobar / rho;
+    s = beta / rho;
+    theta = s * alpha;
+    rhobar = - c * alpha; % msl: I am overwriting a matlab function, maybe change name
```
AnderBiguri, line 119: Agree, any suggestion?

MATLAB/Algorithms/LSQR.m
```diff
@@ -0,0 +1,273 @@
...
+    normr = normr*abs(s);   % ||r_k|| = ||b - A x_k||
+                            % Only exact if we do not have orth. loss 
+    normAtr = phibar * alpha * abs(c); % msl: Do we want this? ||A^T r_k||
+    errorL2(ii)=normr;
+    
+    % (6) Test for convergence. 
+    % msl: I still need to implement this. 
+    % msl: There are suggestions on the original paper. Let's talk about it!
+    
+    if measurequality % msl: what is this??
```
AnderBiguri, line 138: terrible names. Measure_quality computes stuff like MSE, UQI, SSIM, PSNR etc. Its here if you want to evaluate not just the norm, but how much of your image quality measure is changin every iteration. Tends to be off, but can be useful for heuristic stopping criteria testing.

MATLAB/Algorithms/LSQR.m
```diff
@@ -0,0 +1,273 @@
...
+    end
+
+    if ii>1 && errorL2(ii)>errorL2(ii-1)  % msl: not checked
+        % OUT!
+       x=x-alpha*v;
+       if verbose
+          disp(['CGLS stoped in iteration N', num2str(ii),' due to divergence.']) 
+       end
+       return; 
+    end
```
AnderBiguri, line 149: we can remove this maybe, not sure. In CGLS sometimes it would go out, so I stopped. Never sure if it was a math issue or a numerical issue.

MATLAB/Algorithms/LSQR.m
```diff
@@ -0,0 +1,273 @@
+function [x,errorL2,qualMeasOut]= LSQR(proj,geo,angles,niter,varargin)
+
+% LSQR_CBCT solves the CBCT problem using LSQR.
```
AnderBiguri, line 3: LSQR (not LSQR_CBCT). Possibly need to change it in CGLS too. Old names.

MATLAB/Demos/d08_Algorithms03.m
```diff
@@ -63,6 +63,8 @@
  
 % use CGLS
 [imgCGLS, errL2CGLS]=CGLS(noise_projections,geo,angles,60);
+% use LSQR
```
AnderBiguri, line 66: maybe (at a more advanced stage of our work) we can try to explain why use one or the other? Or at least their differences.

AnderBiguri (commented):
Just adding comments to explain and keep track of stuff.

MATLAB/Algorithms/LSQR.m
malena-sabate: So this is not a stopping criterion as such but we are just checking the algorithm is not doing weird stuff? (i.e. loss of orthogonality). We should still have some sanity checks maybe (check if the values of the projected matrix are too small?). ErrorL2 is not the error but the residual norm, right? (or the norm of the discrepancy, value of the fit the data term, etc).

Let's try some examples and see what happens, what do you think?

MATLAB/Demos/d08_Algorithms03.m
malena-sabate: Not sure when you should ever use CGLS! But could be persuaded... we can look into that.

MATLAB/Algorithms/LSQR.m
malena-sabate: actually that is for rho. Change to rho_var or something like that

MATLAB/Algorithms/LSQR.m
AnderBiguri: Exactly, algorithm not doing weird stuff (I mean, it should never increase the residual, right?). Indeed its the residual norm. 

The problem is that I never seen this with synthetic data, only with real data. In real data the system matrix describes the measurement worse, thus these things happen more.

AnderBiguri:
Added python codes. 

TODO for Ander:

- [ ] Add IRN_TV_CGLS to demos, MATLAB
- [ ] Check optional parameter inputs in python
- [ ] Add algorithms to demos in python

MATLAB/Demos/d09_Algorithms04.m
```diff
@@ -191,17 +191,35 @@
                   
 imgSARTTV=SART_TV(noise_projections,geo,angles,50,'TViter',100,'TVlambda',50);           
 
+
+% IRN_TV_CGLS
+%==========================================================================
+%========================================================================== 
+% MALENA, CAN YOU ADD A SMALL TEXT HERE? EXPLAINING THE ALGORITHM IN 3
```
AnderBiguri, line 198: @malena-sabate  can you do this please?

AnderBiguri:
For some reason, IRN_TV_CGLS is 4 times slower in python, investigate....

MATLAB/Algorithms/hybrid_LSQR.m
```diff
@@ -0,0 +1,270 @@
...
+    % Solve the projected problem 
+    % (using the SVD of the small projected matrix)
+    Bk = B(1:ii+1,1:ii);
+    [Uk, Sk, Vk] = svd(Bk);
+    if ii==1
+        Sk = Sk(1,1);
+    else
+        Sk = diag(Sk);
+    end
+    rhsk = proj_rhs(1:ii+1);
```
AnderBiguri, line 118: @malena-sabate `proj_rhs` is never updated, bug?

AnderBiguri:
Mergin this, even if python `hybrid_fLSQR` is not working yet....

Issue #399, opened 2022-09-05
Crash when trying forward projection with flipped head phantom


pjueon:
I'm trying to run `Python/demos/d03_generateData.py` with the head phantom that is flipped upside down.
If I run the original demo code, it works.
But if I add this one line before the forward projection, the program crashes. 
```python
head = np.flip(head, axis=0)  # flip the phantom data upside down
```

I'm not sure but it might be an exception handling issue  between python and Cuda C++.

Or am I missing something?
 
## Expected Behavior
The program doesn't crash.

## Actual Behavior
The program crashes.

## Code to reproduce the problem (If applicable)

```python
import tigre
import numpy as np
from tigre.utilities import sample_loader
from tigre.utilities import CTnoise

geo = tigre.geometry_default(high_resolution=False)

#%% Define angles of projection and load phatom image

# define projection angles (in radians)
angles = np.linspace(0, 2 * np.pi, 50)
# load phatnom image
head = sample_loader.load_head_phantom(geo.nVoxel)


head = np.flip(head, axis=0) # flip the phantom data upside down
print("flip done")

# Simulate forward projection.
projections = tigre.Ax(head, geo, angles)   # crashes here!

print("Ax done")   # not excuted
```

## Specifications 

  - Python version: 3.9.2, with venv.
  - OS: Windows 11
  - CUDA version: Cuda 11.6

AnderBiguri:
I wonder if this has to do with `np.flip()` returning a _view_ and not a _copy_. 

Try to do 

`head = np.flip(head, axis=0).copy() # flip the phantom data upside down
`

pjueon:
@AnderBiguri 
Thanks! That does work. 

One more thing:

I also tried this one
```python
head = np.flip(head.copy(), axis=0) 
```
and it didn't work. Maybe flipped view causes crash(could be a bug)?

pjueon:
Here, casting `img.data` to C++ pointer might be the problem. I'm not familiar with numpy implementation but probably the numpy view's `data` is not flipped and not pointing the position we desire.

https://github.com/CERN/TIGRE/blob/54a854bec2bc5ebf4184c700d6298b31f00c1ec3/Python/tigre/utilities/cuda_interface/_Ax.pyx#L73-L83

pjueon:
Ok after googling about the memory order of numpy, I think I can fix this. I`ll make a PR when I get home.

AnderBiguri:
[...]

PR #400, opened 2022-09-06
Fix numpy memory order issue

pjueon:
- fix for #399 (not only `Ax`, but also all the cuda-numpy interface functions potentially have the same issue.)
- the memory order of the numpy arrays must be c-contiguous
- `numpy.ascontiguousarray` returns a contiguous array. Deep copy only occurs if the memory order is not c-contiguous.

I didn't added it but you can also add warnings when the deep copy occurs by checking `if img.flags['C_CONTIGUOUS']` if you want.

### DIFF
--- a/Python/requirements.txt
+++ b/Python/requirements.txt
@@ -1,5 +1,5 @@
 Cython
 matplotlib
-numpy
+numpy>=1.20.0
 scipy
 setuptools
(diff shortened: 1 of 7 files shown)


pjueon:
@AnderBiguri 
Please let me know your thoughts. 
Adding the warning option might be a good idea I think.

AnderBiguri:
This looks good. I though about the warnings, and while its a good idea, I am not entirely sure where to put them... It should never happen inside algorithms anyway, so maybe we can just forget the warnings, its quite an edge case that the user will not have enough memory.

Issue #402, opened 2022-09-21
Make `computeW`


AnderBiguri:
SART and SART_TV have a different W computation. 

Code is quite repeated, so a new fuunction computeW should be added that all SART-like algorithm use.

PR #403, opened 2022-09-30
Fix bug and typos

twhitbread:
In SART_TV.m, the voxel size used when calculating the W weights is `geo.DSD-geo.DSO`. This errors when the length of `DSD` or `DSO` is greater than 2, i.e. for varying `DSD` and `DSO` (e.g. tomosynthesis scans). This PR changes it to match the voxel size used in SART.m. The rest of the changes are typos/cleaning up a bit.

I will not change the voxel sizes used in SIRT or OS-SART yet, as alignment will come when https://github.com/CERN/TIGRE/issues/402 is implemented. This is purely to fix a bug in the short term.

### DIFF
--- a/MATLAB/Algorithms/OS_SART.m
+++ b/MATLAB/Algorithms/OS_SART.m
@@ -1,34 +1,33 @@
 function [res,errorL2,qualMeasOut]=OS_SART(proj,geo,angles,niter,varargin)
-
 % OS_SART solves Cone Beam CT image reconstruction using Oriented Subsets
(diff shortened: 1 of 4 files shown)


twhitbread:
Also removed `errorL2=[];` from SIRT because it is done twice.

AnderBiguri (approved):

AnderBiguri:
Thanks @twhitbread ! Too many things in my plate, its taking me longer than expected to fix easy bugs. Thanks for helping!

PR #376, opened 2022-06-17
Offset redundancy weighting

grupesh:
Added data redundancy weighting necessary for offset detector geometry (as an option) in all algorithms, except MLEM, tested with simulation & real data..
also includes 
- fix in V matrix for more stability
- additional fix in zeropadding function for geo.COR

PR #415, opened 2022-12-01
Restore control over wang weights in FDK

twhitbread:
Closes https://github.com/CERN/TIGRE/issues/413.

https://github.com/CERN/TIGRE/pull/376 introduced offset weighting for lots of reconstruction algorithms, and tidied up the offset weighting code, but in doing so removed control of the weighting in FDK. In its current state, the decision on whether to do the weighting is made automatically.

In this PR I added the control back in, so the user can turn the weights on/off using the 'wang' input, as before. The default is true. I also removed `theta` from `zeropadding` and inputs `proj` and `geo` from `parse_inputs`, because they are no longer used.

A question remains about lines 47-48: `% Zero-padding to avoid FFT-induced aliasing %TODO: should't this be for all cases, not just wang?`. I don't know the answer to this.


AnderBiguri:
Thanks @twhitbread !

PR #416, opened 2022-12-02
Use original geometry for FDK redundancy weighting

twhitbread:
Following https://github.com/CERN/TIGRE/pull/415 I noticed another bug in FDK offset weighting. It may have been introduced by me in https://github.com/CERN/TIGRE/pull/415, or in the original PR https://github.com/CERN/TIGRE/pull/376.

The `redundancy_weighting` function was using the new `zgeo` structure that comes out of `zeropadding`, which means the detector offset is zero (so the weighting is turned off automatically) and `theta` was being calculated incorrectly anyway. I fixed this by passing in the original `geo` structure as an additional input for FDK only; for all other algorithms there is no zero-padding as far as I'm aware, so the correct `geo` structure is used.

Also fixed a weight bug in MLEM.

### DIFF
--- a/MATLAB/Algorithms/FDK.m
+++ b/MATLAB/Algorithms/FDK.m
@@ -46,12 +46,10 @@ geo.filter=filter;
 if dowang
     % Zero-padding to avoid FFT-induced aliasing %TODO: should't this be
     % for all cases, not just wang?
-    [zproj, zgeo] = zeropadding(proj, geo);
     % Preweighting using Wang function
-    proj=zproj.*redundancy_weighting(zgeo, geo);
-    % Replace original proj and geo
-    % proj = proj_w;
-    geo = zgeo;
+    proj=proj.*redundancy_weighting(geo);
+    [proj, geo] = zeropadding(proj, geo);
+
 end
 
 if size(geo.offDetector,2)==1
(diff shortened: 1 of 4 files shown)


AnderBiguri:
Thanks @twhitbread !
Is there a chance you can give me a geometry to reproduce the bug? The solution is correct but I may try to find a cleaner way of obtaining this behavior than passing 2 geos.

twhitbread:
Sure:

`angles = linspace(0, 2*pi, 101);`
`angles(end) = [];`
`geo = defaultGeometry('nVoxel', [128; 128; 128]);`
`geo.rotDetector = [0; 0; 0];`
`geo.offDetector(1) = 100;`

twhitbread:
I brought in recent changes from `master` and MLEM is failing. Will look for a solution and push to this branch if I can find one.

twhitbread:
The bug is in the computation of `W`. I think this is a mis-named variable: it's usually a forward-projection weight, but in MLEM it was computed by doing a back-projection (so more like V weight?). In the recent PR the code was replaced with the new `computeW` function, which does a forward-projection, so `W` is not the correct size, and we get an error on line 65 of MLEM.

AnderBiguri:
@twhitbread sure!

For reference (for me or you to fix), the `computeV` for MLEM is the same trace as SIRT, the same input params.

twhitbread:
@AnderBiguri using the line from SIRT changes the scale of the reconstruction quite a bit:

image

Is this a problem or are you happy with this?

AnderBiguri:
@twhitbread hum, I must have screwed this in some other way.... Certainly should not happen. I will investigate

PR #418, opened 2022-12-09
Updates to Nikon loader

pjb304:
Using the in built Nikon loader with a scan from Inspect-X 6.8 highlighted the fact that it now stores the calculated CoR in a different field in the .xtekct file, and the reconstructions generated were invalid.  Both issues are fixed in this PR.  The invalid reconstruction issue was due to difference in expected rotation direction.


AnderBiguri:
thanks @pjb304 !

For the sign on the angles: is this a new change in the latest versions of Nikon software, or you think this was wrong for all the cases?

Issue #430, opened 2023-03-04
numpy deprecated np.int in 1.24.0. Find a general solution


AnderBiguri:

tsadakane:
As mentioned in https://stackoverflow.com/questions/21851985/difference-between-np-int-np-int-int-and-np-int-t-in-cython ,
`numpy.int` is an alias for `int`.
```
>>> int is np.int
True
```
```
>>> a=np.int(0x1ffffffff)
>>> print(type(a))
<class 'int'>
<ipython-input-5-7743eb761769>:1: DeprecationWarning: `np.int` is a deprecated alias for the builtin `int`. To silence this warning, use `int` by itself. Doing this will not modify any behavior and is safe. When replacing `np.int`, you may wish to use e.g. `np.int64` or `np.int32` to specify the precision. If you wish to review your current use, check the release note link for additional information.
Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
  a=np.int(0x1ffffffff)
```
I suggest replacing `numpy.int` with `int` as the warning says it does not modify any behaviour.

AnderBiguri:
@tsadakane fantastic, I assumed so, but I didn't do the proper checks. 

I will be busy with other things on the month of March, so TIGRE support will need to wait until April, but if anyone wants to tackle this I'd be happy to check PRs. Otherwise I will just do it in April :)

PR #443, opened 2023-05-16
Fixed np.int in iterative_recon_alg.py

Keyn34:
Minor fix to prevent deprecation-related errors from numpy (#430):

Using the Python installation of TIGRE raised the error:  `AttributeError: module numpy has no attribute int. np.int was a deprecated alias for the builtin int.` This happens while running demos and functions utilizing iterative_recon_alg. Changing `np.int` to `np.int32` fixes the issues, as proposed by the complete numpy error message:
```
AttributeError: module numpy has no attribute int.
np.int was a deprecated alias for the builtin int. To avoid this error in existing code, use int by itself. Doing this will not modify any behavior and is safe. When replacing np.int, you may wish to use e.g. np.int64 or np.int32 to specify the precision. If you wish to review your current use, check the release note link for additional information.
The aliases was originally deprecated in NumPy 1.20; for more details and guidance see the original release note at:
    https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations.
```

This was the only remaining instance of `np.int` I could find.

### DIFF
--- a/Python/tigre/algorithms/iterative_recon_alg.py
+++ b/Python/tigre/algorithms/iterative_recon_alg.py
@@ -380,7 +380,7 @@ class IterativeReconAlg(object):
         :return: None
         """
 
-        ang_index = self.angle_index[iteration].astype(np.int)
+        ang_index = self.angle_index[iteration].astype(np.int32)
 
         self.res += (
             self.lmbda


AnderBiguri:
Thanks a lot :)

PR #476, opened 2023-08-11
Refactor python compilation

tsadakane:
* Made it easy to maintain the CUDA compatibility table
* Fixed the bug that cuda_version is always 11.0

## Test:
Run example.py successfuly
###  Test Environment:
  - Python version: 3.10
  - OS: Windows 10
  - CUDA version: 11.8


AnderBiguri:
Thanks @tsadakane ! Smart solution to the problem

Issue #479, opened 2023-08-16
setup.py install is deprecated


tsadakane:
# Problem
`python setup.py --install` got deprecated. See #478 

1.
While installing with `python setup.py --install` as described in the instruction, the following waring is shown at the beginning:
```
running install
C:\opt\anaconda3\envs\test311\Lib\site-packages\setuptools\_distutils\cmd.py:66: SetuptoolsDeprecationWarning: setup.py install is deprecated.
!!

        ********************************************************************************
        Please avoid running ``setup.py`` directly.
        Instead, use pypa/build, pypa/installer or other
        standards-based tools.

        See https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html for details.
        ********************************************************************************

!!
  self.initialize_options()
C:\opt\anaconda3\envs\test311\Lib\site-packages\setuptools\_distutils\cmd.py:66: EasyInstallDeprecationWarning: easy_install command is deprecated.
!!

        ********************************************************************************
        Please avoid running ``setup.py`` and ``easy_install``.
        Instead, use pypa/build, pypa/installer or other
        standards-based tools.

        See https://github.com/pypa/setuptools/issues/917 for details.
        ********************************************************************************

!!
  self.initialize_options()
running bdist_egg
....
```

2.
After installation finished, `pip list` lists "pytigre" twice. (See https://github.com/CERN/TIGRE/pull/478#issuecomment-1680209701 )

# Possible Solution
According to the article ( https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html ), the alternative for `setup.py install` is `pip install`.

As for the second problem, it is solved by the replacement (after some modification of `setup.py` and `pyproject.toml`)

# Environment
  - python: 3.11 or 3.10+
  - Cython: 3.0.0
  - OS: Windows 10
  - CUDA: 11.8

PR #480, opened 2023-08-17
Use "pip install" for setup

tsadakane:
See #479

### DIFF
--- a/Python/pyproject.toml
+++ b/Python/pyproject.toml
@@ -1,2 +1,5 @@
 [tool.black]
 line-length = 100
+
+[build-system]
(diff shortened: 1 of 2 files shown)


tsadakane:
As far as I know, as for now, there are two options to achieve the entitled goal.
1. Remove `pyproject.toml`
2. Add the two lines below to `pyproject.toml`:
    ```
    [build-system]
    requires = ["setuptools", "wheel", "Cython", "numpy"]
    ```

AnderBiguri:
@tsadakane I am definitely not an expert here, but I think maybe 2) is the way to go? Happy if you think otherwise.

AnderBiguri:
Fantastic. Should we merge this?

PR #508, opened 2023-11-29
pytorch bindings

Emvlt:
Bug appears only in 3D

### DIFF
--- a/Common/CUDA/Siddon_projection.cu
+++ b/Common/CUDA/Siddon_projection.cu
@@ -674,100 +674,90 @@ void computeDeltas_Siddon(Geometry geo,int i, Point3D* uvorigin, Point3D* deltaU
     //End point
     Point3D P,Pu0,Pv0;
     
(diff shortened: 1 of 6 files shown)


AnderBiguri:
Thanks @Emvlt , I'll have a look. AFAIK: 2D works, 3D crashes.

Emvlt:
So, the crash only happens when forward is called a second time. I can compute a first forward pass exactly the way it should be, but the second call always crashes in 3D. The problematic line is 
` x = x.detach().cpu().numpy()`

AnderBiguri:
I am a bit confused on the different way you tread the 3D version and the 2D version, can you explain further?

In TIGRE, there is no 2D operators, there are only 3D operators that can take 3 rd dimension == 1, so a priori, there is no need to change anything in the way you call  Ax(). 

Would it not make it easier to treat these the same way? i.e. always assume you get a 3D tensor (+batch) as input. This makes it 4D for the THREE_D image case right? why do you have 5D sometimes in the THREE_D case?

Emvlt:
This pull request accounts for:

- Forward and Backward Operators as torch.nn.Modules
- Demo 23: CT reconstruction by iterative data discrepancy minimisation (there is a TODO for @AnderBiguri in this file + there is a known bug for 3D)
- Demo 24: LPD training on MNIST data

Issue #383, opened 2022-07-18
Convert ASTRA geometry to TIGRE geometry? 


sjmgarnier:
Hello! This is more a question than an issue. Is there a way to convert ASTRA geometries into TIGRE geometries easily? Thanks in advance for your help.

AnderBiguri:
Hi! there isn't a function to do so (if you find one, please do commit it to TIGRE!), but unless you have a particularly weird geometry, it should be easy to create it or just fill the geometries by hand using ASTRA values. 
Give it a shot, I also have a little experience with ASTRA so I may be able to help if you get stuck.

sjmgarnier:
@AnderBiguri Thanks for your very quick reply! We do have a bit of a weird geometry, that's why I'm looking for something that'll do the work for me :-) Long story short, we're using a "homemade" CT scanner in semi-fieldwork conditions in the tropical forest (you can read about an earlier work of ours here if that is of interest to you: https://arxiv.org/abs/2110.09017). We can't guarantee that the axis of rotation is always perfectly aligned with the source-detector axis, so we rely on posthoc calibration with fiducial markers for estimating the position of the source-detector couple relative to the axis of rotation. As a result, we have a different configuration for each angle.

AnderBiguri:
@sjmgarnier Oh, that is quite interesting! 

Depends on how you obtain this parameters, adding them directly to TIGRE may be fairly easy. Most parameters (except detector/image sizes) can be given per-projection as an array in TIGRE, so given a variable DSD / DSO you can just input it as an array. Similarly, uCT machines tend to have a Axis/center of rotation (COR) error, so as you say, they are posthoc estimated and corrected for that. You can do that with the (somehow redundant but convenient) geometry parameter geo.cor. It basically translates to a horizontal shift on the axis of rotation from the source-detector line. This parameter can also be an array, per projection. 

I think adding these in ASTRA can be a bit more complicated, so maybe your best go is to directly put them into TIGRE from whichever code you have that estimates them, as TIGRE geometry is slightly more user intuitive (for a standard scan, at least).

sjmgarnier:
@AnderBiguri Thanks, I'll look into that direction. We're using the standard `estimateWorldCameraPose` from Matlab for the calibration.

AnderBiguri:
[...]

PR #515, opened 2024-01-19
Twin Robotic Computed Tomography Geometries / ASTRA

wittlsn:
Hello TIGRE team! 
I am a PhD student at the Deggendorf Institute of Technology where we are working with a twin robotic computed tomography scanner. I usually use the astra toolbox for reconstruction, but I wanted to try out the many algorithms implemented in the TIGRE toolbox and the cool features like the multi GPU! I found it rather unintuitive to convert the complex arbitrary geometry into TIGRE, so I've added a utility function. I tested the results and made a short colab example with the conversion from astra to TIGRE (I thought it might be interesting for others).

https://colab.research.google.com/drive/1c6tEySsv63zIdkHvo6agg3O2vJoKYTC1?usp=sharing

With our twin robot CT system we have correction values for each projection. I found the TIGRE interface rather unintuitive, especially compared to the Astra 'cone_vec' interface. I took a quick look at the TIGRE cuda files and found that in the forward and backprojector the interface is more or less the same as the Astra 'cone_vec' geometry. Is it worth looking into this and maybe making a more "arbitrary" python interface for TIGRE? I might look into it when I have some spare time.

Greetings Simon

### DIFF
--- a/Python/tigre/utilities/common_geometry.py
+++ b/Python/tigre/utilities/common_geometry.py
@@ -158,6 +158,9 @@ def ArbitrarySourceDetectorFixedObject(
             focal_spot_position_mm, detector_center_position_mm)
         focal_spot_position_mm = focal_spot_position_mm - trajectory_center_mm
         detector_center_position_mm = detector_center_position_mm - trajectory_center_mm
+    else:
+        trajectory_center_mm = np.zeros((3,))
+        
     
     geometry = Geometry()
     geometry.mode = 'cone'
--- a/Python/tigre/utilities/common_geometry.py
+++ b/Python/tigre/utilities/common_geometry.py
@@ -138,20 +138,31 @@ def ArbitrarySourceDetMoveGeo(geo,s_pos,d_pos=None,d_rot=None) -> Geometry:
 
 
 def ArbitrarySourceDetectorFixedObject(
+        geometry: Geometry,
         focal_spot_position_mm: np.ndarray, 
         detector_center_position_mm: np.ndarray, 
         detector_line_direction: np.ndarray, 
         detector_column_direction: np.ndarray,
+        origin_mm: np.ndarray | None = None,
         use_center_correction: bool = True) -> Geometry:
     """
-    geo: Geometry, that gets appended
-    ...
-    detector_rotation_marix: x_axis: (0, 0) -> (0, 1);  y_ais (0, 0) -> (1, 0)
+    geo: Geometry object
+    focal_spot_position_mm: position of the source, 
+    detector_center_position_mm: position of the detector center, 
+    detector_line_direction: detecor line vector from pixel (0, 0) -> (0, 1), 
+    detector_column_direction: detecor column vector from pixel (0, 0) -> (1, 0),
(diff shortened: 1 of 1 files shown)


AnderBiguri:
Hi Simon, 

Thanks, this is fantastic work, and indeed super useful. #383 mentions this, because indeed, if you have a standard CBCT, TIGRE geometry is more intuitive, but with arbitrary motions TIGRE becomes quite complex to define, and ASTRA is quite good for that.

I am correct then assuming that your function ` ArbitrarySourceDetectorFixedObject` is essentially equivalent to `from_ASTRA_geo_to_TIGRE_geo()`?

AnderBiguri:
> the `ArbitrarySourceDetectorFixedObject` function allows reconstruction with the astra geometries 

This is wonderful!

>  This is why the volumes can vary with a constant offset.

What do you mean here?

> I just had a quick look at the cuda files, but it seems that in the projector the 3D position of the source and detector is needed?

That parameter you are linking there is the location of the origin of the detector in "projection geometry". Basically encoding the detector offset. Its the world coordinates in "projection geometry" for pixel (0,0) in the detector. Note that the "projection geometry" is this slightly strange geometry where the first voxel of the volume is always at (0,0,0) and the voxel sizes are always (1,1,1). The entire rest of the world is rotated and scaled accordingly. It makes the main loop at L193 much more efficient. 

> So I thought it might be worth looking into whether an interface can be made for this geometry / data?

To have an interface in this "projection geometry" may be possible, but no sure if desirable, as its a bit complex and not human readable easily. However, this geometry is computed from the TIGRE geometry at:
https://github.com/CERN/TIGRE/blob/master/Common/CUDA/ray_interpolated_projection.cu#L632
So it would be possible to change the type of geometry that the CUDA code takes in. However I think that the best way to do so is still to do it directly in python, as you did in this PR.

wittlsn:
> What do you mean here?

We do not know exactly the origin of our CT trajectory because we use a pose estimation algorithm and get the projection geometries relative to a phantom. So I implemented this function:

https://github.com/wittlsn/TIGRE/blob/roboct_thd/Python/tigre/utilities/common_geometry.py#L301

If the origin is known, the source and detector positions should be transformed before using the function. Then the volumes should be reconstructed on the same volume grid. 

> So it would be possible to change the type of geometry that the CUDA code takes in. However I think that the best way to do so is still to do it directly in python, as you did in this PR.

OK!

AnderBiguri:
> If the origin is known, the source and detector positions should be transformed before using the function

Would it make perhaps more sense to have as an input `trajectory_center_mm` (I assume this is the location of the axis of rotation?) with a default value that is the most common (e.g. (0,0,0)). Then have the possibility for the user to input either a value, or `None`. If a value is input, then the correction happens inside. If `None`, your function is called. 

This would keep a functionality (correcting for trajectory center) locate in a single place, making it hopefully more intuitive to use. What do you think?

AnderBiguri:
Thanks! Are you OK with me merging this?

wittlsn:
Yes! Thanks for the quick replies! I have also updated the colab example.

AnderBiguri:
Fantastic :) And thanks again for contributing, TIGRE keeps alive thanks of people like you :)

PR #527, opened 2024-03-15
Add demos/d19_DBTexample.py

timcogan:
Add `d19_DBTexample.py` which is adapted from `d19_DBTexample.m`.


AnderBiguri:
Super good @timcogan , thanks a lot!

The only comment I have is: would you mind reformating it so it doesnt have a `main` function? I understand is quite typical python, but its for uniformity purposes, comparing the other demos. 
Thanks!

AnderBiguri:
Thanks! I'll give it a test and merge soon.

Issue #523, opened 2024-03-08
Make installation easier


AnderBiguri:
Can we automatically find the compiler/MVS version, and just rename the xml file ourselves? Ideally we don't want users to do anything other than just run the compilation script. 

Can we even provide pre-compiled libraries? I think this is harder to do without loss of generality.

tsadakane:
Can we use the following snippet to get the Visual Studio version?
```
% path to vswhere
path_to_vswhere = "c:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe";

% Run vswhere to get version string
if exist(path_to_vswhere, 'file') == 2
    [~, vs_version_string] = system('"'+path_to_vswhere+'" -latest -property installationVersion');
    fprintf('Visual Studio version: %s\n', vs_version_string);
else
    error('Error: vswhere.exe not found');
end

% Split version string
vs_versions = strsplit(vs_version_string, ".");
vs_major_version = vs_versions{1, 1};
fprintf('vs_major_version: %s\n', vs_major_version);

if strcmp("17", vs_major_version)
    fprintf("2022\n");
elseif strcmp("16", vs_major_version)
    fprintf("2019\n");
elseif strcmp("15", vs_major_version)
    fprintf("2017\n");
elseif strcmp("14", vs_major_version)
    fprintf("2015\n");
else
    error("Unknwon version %s\n", vs_major_version)
end
```

This worked on my environment where VS2022 and2019 are installed ("2022" is printed).

PR #528, opened 2024-03-18
Enh make installation easier 

tsadakane:
A solution for #523

The users only have to run the compilation script `Compile.m`.  It detects the version of the latest visual studio, copies the appropriate xml file to "mex_CUDA_win64.xml", and starts compilation.

Matlab, windows.

Tested on a system that Visual studio 2019, 2022 and Matlab 2022a are installed.


AnderBiguri:
Looks good!

I haven't tried, but I wonder what happens if vscode is not in `path_to_vswhere` . The XML does try to find it in some other places, I think (e.g. envvars)? Maybe I need to test it in a machine with non-standard vscode install location to double check.

tsadakane:
Hi, @AnderBiguri ,
Thank you for your reply.

> what happens if vscode is not in `path_to_vswhere`

`vswhere` is an utility that detects Visual Studio instances.
According to [this page](https://learn.microsoft.com/en-us/visualstudio/install/tools-for-managing-visual-studio-instances?view=vs-2022), 
```
vswhere.exe is automatically included in Visual Studio 2017 and later, 
or you can download it from [the vswhere releases page](https://github.com/Microsoft/vswhere/releases)
```
If `vswhere` was not found in `Compile.m`,  it dies at L55.

I think there should be an option to specify the version of the Visual Studio, and I'd like to make additional modification to this PR.

AnderBiguri:
Didn't know that! Thanks for explaining :) So I assume vswhere is always installed in the same place, regardless of user choice in installation locations? If that is true, then super nice tool!

Sure, do let me know when its ready for review 👍

tsadakane:
@AnderBiguri , 
Please review my PR.

To compile using the latest Visual Studio installed, just
```
>> Compile
```
This detects the latest installed visual studio and copies the appropriate xml file to `mex_CUDA_win64.xml`.

To compile using MVS 2013 even if MVS2022 is installed, for example, 
```
>> Compile -mex_xml mex_CUDA_win64_MVS2013.xml
```
when `-mex_xml` option is set, the specified xml file is copied to `mex_CUDA_win64.xml` and is used for the setup.
If the specified file was not found, `Compile` dies.

In the both cases above, `mex_CUDA_win64.xml` will be overwritten if it exists before `Compile` is run.


Note:
Test env.
* Matlab 2022a
* Windows 10
* Visual Studio 2022, 20199 are installed
* CUDA 11.8

AnderBiguri:
Looks really good!

We should also add this to the install instructions README. I will test later today just for sanity check, and then we can merge.

tsadakane:
@AnderBiguri , 
`Compile.m` is updated to support VSs which do not have `vswhere`.

I'd like to refrain from editing README/Instructions due to my English language skills.
I can suggest the following:
* From "Simple Instructions", you can remove the step 2.
* From "Step by Step Instructions", you can remove the step 7 (and maybe step 6?)

AnderBiguri:
Hi @tsadakane, yes apologies, I was going to do it myself, just life got in the way. 
Thanks for the hard work. I'll do a commit to this branch with the updated readme and then I will merge.

Issue #542, opened 2024-05-07
UBUNTU met "UnboundLocalError: local variable 'test_data' referenced before assignment" when using "demo/d03" test if it's installing


GreameLee:
- python version: 3.7
  - OS:ubuntu 18.4
  - CUDA version:12.2

AnderBiguri:
hi @GreameLee , can yo tell me the entire error (incuding line)?

GreameLee:
Traceback (most recent call last):
  File "d03_generateData.py", line 35, in <module>
    head = sample_loader.load_head_phantom(geo.nVoxel)
  File "/home/user/anaconda3/envs/CT/lib/python3.7/site-packages/tigre/utilities/sample_loader.py", line 27, in load_head_phantom
    image = test_data["img"].transpose(2, 1, 0).copy()
UnboundLocalError: local variable 'test_data' referenced before assignment

GreameLee:
> hi @GreameLee , can yo tell me the entire error (incuding line)?

Traceback (most recent call last):
  File "d03_generateData.py", line 35, in <module>
    head = sample_loader.load_head_phantom(geo.nVoxel)
  File "/home/user/anaconda3/envs/CT/lib/python3.7/site-packages/tigre/utilities/sample_loader.py", line 27, in load_head_phantom
    image = test_data["img"].transpose(2, 1, 0).copy()
UnboundLocalError: local variable 'test_data' referenced before assignment

AnderBiguri:
@GreameLee have you installed the toolbox? seems that the data is missing, but this should only happens if you did not `pip install .` as the instructions suggest. Or did you do that, but still get this error??

GreameLee:
@AnderBiguri YesI  do all instructions step by step and successfully do it and try several times but still get this error. And when I go to the  sample_loader.py file and put test_data as an absolute path of head.mat, it can run now. I guess there is a problem with the relative path.

AnderBiguri:
Thanks! for your own, that change should suffice to keep you going. I will investigate soon to see if I can find a fix that is general.

PR #545, opened 2024-05-18
solved the issue542

immortalmin:
Wrong path for head.mat


AnderBiguri:
Thanks a lot for fixing this! fix #542

Issue #544, opened 2024-05-13
Inaccurate FDK reconstruction for low voxel size


simonCtBern:
I'm trying to use tigre for reconstruction cone-beam CT images with variable scan geometries. 
However, I encountered unexpected results for a test data set with simulated projections of a simple circular scan trajectory. 

## Expected Behavior

The quality of the reconstructed volume should not depend on "linear scaling of the object / projection magnification". 

## Actual Behavior

Reconstruction of the simulated scan using FDK gives good results for the following set of geometry parameters: 
- Distance from source to detector (DSD) = 556.0 mm
- Distance from source to origin (DSO)= 200.0 mm 
- Size of each voxel (dVoxel) = [0.07194245, ...] mm

However, reconstruction of the same data provides poor results if a larger magfinication is provided as geometry parameters:
- Distance from source to detector (DSD) = 556.0 mm
- Distance from source to origin (DSO)= 6.0 mm 
- Size of each voxel (dVoxel) = [0.00215827, ...] mm

All other parameters (number of pixels, pixel size) and the projections are identical for both reconstructions:
- Number of pixels (nDetector) = [1200 2000]
- Size of each pixel (dDetector) = [0.2 0.2] mm
- Number of voxels (nVoxel) = [1200 2000 2000]

I tried changing the accuracy parameter of the geometry over a wide range of several orders of magnitudes without visible effect. 

It seems that the is some "absolute numerical accuracy threshold" which creates problems at low voxel size? 
Is there a way around this?
(Typing a wrong magnification as a workaround may not matter for an ideal circular scan trajectory, but my goal is to test variations/deviations in the geometry for all projections individually, which would be influenced by the magnification.)

## Scan setup

The test data for the reconstruction is a simulated circular cone-beam scan of a hexagonal object with a hole in the center (roughly representing a hexagonal nut). 

result with large DSO (low magnification):

![tigre_large_SRD](https://github.com/CERN/TIGRE/assets/169674527/b3d7b847-7934-4d04-b18f-49ca0eb3e5e2)

result with low DSO (high magnification), "correct" simulation setup with bad reconstruction:

![tigre_nominal_SRD](https://github.com/CERN/TIGRE/assets/169674527/72aadf86-f182-4898-ba69-703b7194dab1)

## Specifications 

  - python version: 3.10.14
  - OS: Windows 10 64 bit
  - CUDA version: 11.7
  - GPU: 2x GeForce GTX TITAN X (12 GB), compute capability 5.2

[...]

Issue #355, opened 2022-01-27
Tiny symmetry mismatch, unsure where it comes from. 


AnderBiguri:
## Expected Behavior

Projections of a symetric phantom need to produce a symetric projection (for certain geometries)

## Actual Behavior

A tiny tiny tiny numerical offset exists. Much smaller than half of a pixel.

## Code to reproduce the problem (If applicable)

```
     geo=defaultGeometry('nDetector',[512,512],'nVoxel',[512,512,512]);
    geo.dDetector=[0.280;0.280]*4;
    geo.sDetector=geo.dDetector.*geo.nDetector;
    geo.dVoxel=[1,1,1]';
    geo.sVoxel=geo.nVoxel.*geo.dVoxel;
    
    geo.DSO=600;
    geo.DSD=900;


        sphere=@(x,y,z,r,x0,y0,z0)((x-x0).^2+(y-y0).^2+(z-z0).^2<=r.^2);
        n_spheres_per_plane=6;
        
        % define locations of spheres and radious (in pixel units)
        r_sphere=5;
        l=100*geo.dVoxel(1); %for later
        z0=[100+geo.nVoxel(3)/2 -100+geo.nVoxel(3)/2+1]; %positive/negative
        
        t = linspace(0,2*pi,n_spheres_per_plane+1); t(end)=[];
        x0 = r.*cos(t)+geo.nVoxel(1)/2+0.5;
        y0 = r.*sin(t)+geo.nVoxel(2)/2+0.5;
        
        % get coordinates of all pixels
        [x,y,z]=meshgrid(1:geo.nVoxel(1),1:geo.nVoxel(2),1:geo.nVoxel(3));
        
        % crate all spheres
        sample=zeros(geo.nVoxel','single');
        for ii=1:n_spheres_per_plane
            sample=sample + sphere(x,y,z,r_sphere,x0(ii),y0(ii),z0(1))*(1+single(ii==1));
            sample=sample + sphere(x,y,z,r_sphere,x0(ii),y0(ii),z0(2))*(1+single(ii==1));
        end
    end


%%
        geo.offDetector=[30;0].*geo.dDetector;
        angles=0;
        proj=Ax(sample,geo,angles,'interpolated'); % happens for both projectors

       % symmetry check:
       imshow([proj(1:end/2,:,1)-flipud(proj(end/2+1:end,:,1))],[]);colorbar
       % notice the error is a shift upwards of the sample. its about 0.5% of error in pixel value. 
[...]

PR #555, opened 2024-06-05
float precision fix for issue #544

simonCtBern:
Changed the function computeDeltasCube() in voxel_backprojection.cu to double precision, including double precision versions of coordinate rotation functions. Added a double precision Point3Ddouble struct to types_TIGRE.hpp. Integrated the computeDeltas_Siddon() function in Siddon_projection.cu from [IbexInnovations](https://github.com/IbexInnovations/TIGRE/blob/8d8c8eb4e7be922a69512baa08b4dad5bf516cc3/Common/CUDA/Siddon_projection.cu#L680) related to issue https://github.com/CERN/TIGRE/issues/355.

### DIFF
--- a/Common/CUDA/Siddon_projection.cu
+++ b/Common/CUDA/Siddon_projection.cu
@@ -674,100 +674,90 @@ void computeDeltas_Siddon(Geometry geo,int i, Point3D* uvorigin, Point3D* deltaU
     //End point
     Point3D P,Pu0,Pv0;
     
-    P.x  =-(geo.DSD[i]-geo.DSO[i]);   P.y  = geo.dDetecU*(0-((float)geo.nDetecU/2)+0.5);       P.z  = geo.dDetecV*(((float)geo.nDetecV/2)-0.5-0);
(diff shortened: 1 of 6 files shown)


AnderBiguri:
Thanks a lot!
I think this would also need to be propagated to at least `voxel_backprojection2.cu`. Are you happy to do it or should I do it?

simonCtBern:
Yes I wanted to comment that I only copied the changes to voxel_backprojection2.cu, but I don't know where this function is called, so I couldn't test it. 
Also, I simply copied the coordinate rotation functions to voxel_backprojection2.cu since I don't really know the structure of the code, so they are defined identically in 2 places now. You'll probably want to clean this up...

AnderBiguri:
hi @simonCtBern , I have done a commit to this PR, do you mind double checking the code for sanity? Just run whatever experiment this was fixing, just to make sure I haven't broken it :) 
Otherwise it runs on my PC locally with the same behaviour as before my commit.

simonCtBern:
@AnderBiguri Looks good, the inaccuracy is still gone after pulling and recompiling.

PR #556, opened 2024-06-08
Issue410: Added groundtruth comparion flags to ./python/TIGRE/algorithms

ameyakhot:
Submitting my first every PR. Please be free to share feedback if there was anything that needs to be fixed. Thank you for the opportunity.

email: ameyakhot18@gmail.com

### DIFF
--- a/Python/tigre/algorithms/iterative_recon_alg.py
+++ b/Python/tigre/algorithms/iterative_recon_alg.py
@@ -212,6 +212,9 @@ class IterativeReconAlg(object):
             self.Quameasopts = (
                 [self.Quameasopts] if isinstance(self.Quameasopts, str) else self.Quameasopts
             )
+            # appending error_norm is groundtruth is Null
+            if self.groundtruth is not None:
+                self.Quameasopts.append('error_norm')
             setattr(self, "lq", np.zeros([len(self.Quameasopts), niter]))  # quameasoptslist
         else:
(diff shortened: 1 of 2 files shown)


AnderBiguri:
Thanks @ameyakhot !

This is a good step towards fixing the issue, but perhaps the issue wasn't clear enough, my bad!

This helps start with it, but now we need to do something with that groundtruth :) 

This is related to the `Quameasopts` variable, that sets the quality measures options.  As it is now, each iteration will check the quality measures against the previous iteration. The idea of adding a groundtruth is that when this is given (not otherwise), the quality measures should be compared against the groundtruth, instead of comparing them against the previous iteration.

ameyakhot:
I see. So for every iteration, if the `Quameasopts` variable needs to be compared to groundtruth (if given else None). What actions should be taken if they `Quameasopts` is equal to groundtruth? Can you share some details? 

Thank you for your support.

AnderBiguri:
hi @ameyakhot , no not the `Quameasopts` that is the options variable that is used to control that. 

I would suggest having a quick look on how this variable is used in the code. You can also check the MATLAB code (e.g. SART.m) to see how MATLAB handles this.

ameyakhot:
Thank you for the explanation @AnderBiguri . I'll take a look for sure at the matlab and see how it works and try to understand what needs to be done in here.

AnderBiguri:
Closing because this seems stale, feel free to reopen if you decide to keep working on it.

PR #563, opened 2024-07-01
pytorch bindings with CPU/GPU pingpong + associated demo

Emvlt:
Updated pull request for the Pytorch bindings of TIGRE


AnderBiguri:
Trying this again! We had a great realization in Sweeden in June, and I forgot absolutely everything about it. I can't find my notes nor the code we changed. Do you remember anything?

AnderBiguri:
Would you be able to test something for me?

In the new demo script, I set the GPU:

https://github.com/CERN/TIGRE/blob/f89a0fff1e64558ec9cd2308653f12bdf29f3ee2/Python/demos/d25_Pytorch.py#L186-L190

if pytorch and TIGRE are in the same GPU, I get a crash. The code now sets TIGRE to GPU 3 and pytorch to 2, but if I change theline of pytorch GPU to 
```
PYTORCH_GPU_ID =3 
```
Then I get an invalid argument error of CUDA. Can you see if this is something my side, or you can also test?

PR #619, opened 2024-12-10
Fix bug related to ComputeV for PICCS algorithm

Pooryamn:
Hi @AnderBiguri 

I encountered an issue with the new version of the ComputeV function while working with TIGRE and running the PICCS algorithm. The new function works well for ASD_POCS but not for PICCS. To resolve this, I used the body of the old ComputeV function with the new function definition. This modification seems to work for all algorithms, including PICCS.

Could you please review the changes and verify if they are suitable for the repository?

Thank you!

### DIFF
--- a/MATLAB/Algorithms/PICCS.m
+++ b/MATLAB/Algorithms/PICCS.m
@@ -135,7 +135,7 @@ while ~stop_criteria %POCS
         %         weigth_backprj=bsxfun(@times,1./V(:,:,jj),backprj); %                 V * At * W^-1 * (b-Ax)
         %         f=f+beta*weigth_backprj;                          % x= x + lambda * V * At * W^-1 * (b-Ax)
         % Enforce positivity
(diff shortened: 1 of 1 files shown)


AnderBiguri:
Hi @Pooryamn Thanks for looking at this! indeed it requires some fixing. 
while this may work numerically, I think its mathematically incorrect, unfortunately. This computes an impact weight of the backprojection (\sum A^T*I), meaning its very important that `dVoxel` does not change mathematically speaking. As this needs to be used with an image, `nVoxel` also can't change, ultimately making also `sVoxel` not allowed to be changed. 

Theoretically this "expansion" of the projections is not needed, but I found out numerically that sometimes it causes edge cases in edge voxels ti have very small values, wich causes issues with numerical division, which is why I expand it. It still introduces many errors at certain geometries (but not others). I wonder if expanding it further may fix them. 

But, in any case, we can't change the volume if we want to be mathematically accurate.

PR #643, opened 2025-03-14
Move dependencies into pyproject.toml

purepani:
In an effort to be more compliant with modern python packaging standards, I moved the information previously defined in `requirements.txt`, `requirements_dev.txt`, and some of `setup.py`. `pip install .` should just work in the Python directory, assuming valid cudatoolkit and gcc installations.

There's more that can be done to simplify this more(like including cuda-toolkit-nvcc from pypi directly instead of requiring the user to install it), but that seemed a bit more involved, so I decided to start with just this for now.

I also including the [uv](https://docs.astral.sh/uv/) lock file, since that's the tool I normally use for python project management(and in order to encourage you to try out the tool). I can remove it if you don't feel like learning new tools but I figured I'd at least pitch it :). I didn't include instructions in the commit for it just for simplicity, but once you install it, you can just go into the folder and run `uv run example.py` and it should just work


AnderBiguri:
Thanks! I am away now, but I'll test ASAP and merge it :)

purepani:
I moved `setup.py`, `pyproject.toml`, and `uv.lock` to the top level. I did this so that `setup.py` can refer to `Common/` without using relative paths. 
This is the easiest way for `uv` to build wheels correctly, since it builds an sdist first and doesn't include the relative path for "Common" by default in the sdist. There might be a way to get setuptools to include the directory in the sdist, but I didn't figure it out.

All this changes is that you run any python commands from the top level of the project instead of in the Python folder.
Additionally, `uv build` successfully builds a wheel as long as `CUDA_PATH` is set. 

Note that I also changed the linux library path from `lib64` to `lib`; when I installed `cudatoolkit`, `lib64` didn't seem to exist, so I think this makes more sense.

AnderBiguri:
Great! thanks a lot. 

So if I understand correctly, this is all just tidying up files to be more compliant with standard python libraries, correct?

Mr-Blue-Sky-Candy:
Hello, I was also planning to do this contribution. Thanks @purepani. This change would also make publishing to https://pypi.org/ a lot easier, which would make installation as simple as `pip install tigre`

Issue #131, opened 2019-07-04
Convert CUDA into HIP for AMD GPUs


AnderBiguri:
This code seems to convert CUDA into HIP, that runs on AMD GPUs.

https://github.com/ROCm-Developer-Tools/HIP

AnderBiguri:
It's unlikely that I will tackle this myself, leaving it open just to see if anyone would want to do so....

PR #646, opened 2025-03-20
HIP interface

purepani:
I've successfully compiled the project to HIP*! Addresses https://github.com/CERN/TIGRE/issues/131.
*Currently only compiles to amd, haven't tested it(am currently running one of the demos which is taking a while), and no way to switch between nvidia backend and amd backend for compilation atm, and only on the Python backend right now, so definitely not ready to be merged.

It does compile successfully though!
How long are the demos supposed to take on a reasonably powerful GPU(say a 3080)? I've been leaving `demos/d16_2Dtomography.py` running on my 7800 XT for ~10-15 minutes now with my GPU pinned at 100%(so I assume it's doing something :smile:)


purepani:
Note: all cpp and cu files were autogenerated using `hipify`, aside from some small fixes needed to get it to work. 
One of those fixes will technically change the numbers slightly: for `__frcp_rd` and `__fsqrt_rd`, the corresponding HIP api doesn't exist, so they were switched to `__frcp_rn` and `__fsqrt_rn`(both of them are switched from round down to round to nearest even mode).

purepani:
I left it on overnight and it still didn't finish 
Hmm

AnderBiguri:
@purepani sorry, I am away and with no access to computers now. Thanks for this PR! but indeed, the demos should be quite short, at they should tell you how long they will run for (the algorithms) within a minute or so.

purepani:
Yeah that demo said it estimated 9 hours(and running it overnight was definitely longer than that too).

Anyway, I might not have time to get back to this PR in particular, but someone else can pick this up as a starting point to finish this.

AnderBiguri:
That 9h is definetly something wrong, demos should finish fairly quick....

AnderBiguri:
This is superseeded by #747

PR #650, opened 2025-04-01
Modify and update the code related to the PICCS algorithm module.

coldly01:
Fixed bugs in the existing PICCS algorithm CUDA files and added additional code required to invoke the algorithm.

### DIFF
--- a/Common/CUDA/PICCS.hpp
+++ b/Common/CUDA/PICCS.hpp
@@ -55,7 +55,7 @@ Codes  : https://github.com/CERN/TIGRE
 #include "TIGRE_common.hpp"
 #include "GpuIds.hpp"
 
-void piccs_tv(const float* img,const float* prior, float* dst,float alpha, float ratio, const long* image_size, int maxIter, const GpuIds& gpuids);
+void piccs_tv(float* img,float* prior, float* dst,float alpha, float ratio, const long* image_size, int maxIter, const GpuIds& gpuids);
 
 
 #endif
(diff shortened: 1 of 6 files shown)


Common/CUDA/PICCS.hpp
```diff
@@ -55,7 +55,7 @@ Codes  : https://github.com/CERN/TIGRE
 #include "TIGRE_common.hpp"
 #include "GpuIds.hpp"
 
-void piccs_tv(const float* img,const float* prior, float* dst,float alpha, float ratio, const long* image_size, int maxIter, const GpuIds& gpuids);
+void piccs_tv(float* img,float* prior, float* dst,float alpha, float ratio, const long* image_size, int maxIter, const GpuIds& gpuids);
```
AnderBiguri, line 58: Why did you remove `const` here?

Python/tigre/algorithms/pocs_algorithms.py
```diff
@@ -128,6 +130,10 @@ def __init__(self, proj, geo, angles, niter, **kwargs):
     def run_main_iter(self):
         stop_criteria = False
         n_iter = 0
+        if (self.regularization=="PICCS"):
+            # Modify the `set_res` function in the `Python\tigre\algorithms\iterative_recon_alg.py` file as needed.
+            res_prior = copy.deepcopy(self.res)
```
AnderBiguri, line 135: hum not sure about this. `res` is the current variable right? You may want to initialize the algorithm to one image, but have a prior of another, so I think there should be a new kwarg for `prior`

AnderBiguri (changes requested):
Thanks a lot! super useful. 

Can you please answer the questions, so I understand better?

Common/CUDA/PICCS.hpp
AnderBiguri: @coldly01  makes sense :) can you put it back then?

PR #653, opened 2025-04-16
Update pyproject.toml python version to <3.12 instead of <=3.11

DannyRuijters:
For some reason my docker build failes on python version <=3.11 (I am using python 3.11.12), while it succeeds with python version to <3.12


purepani:
Oh yeah this is the correct way to do it. I think <=3.11 means <=3.11.0.

purepani (approved):
LGTM

PR #663, opened 2025-06-04
added "offSource" to adjust the position of X-ray source

defThu:
Hello, I am a Phd student in Tsinghua University, thank you very much for open-sourcing the TIGRE program. I've been using it for quite some time. Previously I noticed that in the .cu code, apart from the parallel beam programs, the X-ray source positions were actually configured, but except for S.x all other positions were defaulted to zero. 

To better match the actual situation where the radiation source might be offset or simulate the oversampling of focal spot, I modified the underlying program by adding the "offSource" parameter. 

After my own checks, no obvious bugs were found. I hope this modification can be merged into the master branch, allowing this open-source program to have broader applications.

Best Regards,
Yifan Deng
Tsinghua University

### DIFF
--- a/MATLAB/Utilities/computeV.m
+++ b/MATLAB/Utilities/computeV.m
@@ -25,63 +25,6 @@ if ~isfield(geo,'mode')||~strcmp(geo.mode,'parallel')
         auxgeo.DSO = geo.DSO(auxindex);
         auxgeo.offOrigin = geo.offOrigin(:,auxindex);
         auxgeo.offDetector = geo.offDetector(:,auxindex);
-        function V=computeV(geo,angles,alphablocks,orig_index,varargin)
-if nargin>=6
-    [gpuids] = parse_inputs(varargin{1:length(varargin)});
-elseif nargin==4
-    gpuids = GpuIds();
-else
-    error('Wrong amount of inputs, 4 or 6 expected');
-end
-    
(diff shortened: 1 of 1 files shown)


defThu:
Thank you for your quick and detailed response. Your reply helped me better understand the program and the issues with my modifications. Regarding the questions above:
1. I may misunderstand the geometry in backprojection code, now I agree that the backprojection program cannot be modified so simply. 
2. Right, here I used the modified code for the simulation of 1) forward projection when the ray source has an offset. 2) The simulation of the penumbra effect by the oversampling of focal spot and then the backprojection without source offset. I think this modification can be useful for the two case.
3. However, my confusion is that, if I want to simulate an offset X-ray source that is not aligned with the rotation center, whether the modification of `geo.offDetector` and `geo.offOrigin` is enough? I think `geo.offDetector` and `geo.offOrigin` are used to add the offset between the center of detector/phantom and the rotation center. This is also why I wanted to add the offSource parameter.

defThu:
Hi @AnderBiguri, Thank you for your reply. I think I understand what you mean. I previously misunderstood that "offOrigin" only moves the object without moving the rotation center. However, if both the object and the rotation center are moved at the same time, then "offSource" is equivalent to directly changing "offDetector" and "offOrigin."

defThu:
Hi @AnderBiguri , I made a new pull request as I did some new modifications of the backprojection code, and the reconstruction images with source offset look good.

PR #666, opened 2025-06-06
added "offSource" to adjust the position of X-ray source

defThu:
Hi AnderBiguri, I added a new pull request here as I further modified the .cu code of backprojection and fixed some bugs in the last pull request.

I checked the modified code and added a demo named "d025_OffsetSource" in the demo dir. The results of the demo looks good. For example, the difference image between the projection with and without source offset is:
difference_of_proj_with_source_offset

And the reconstruction images are similar:( Left: reference image without source offset; Right: reconstruction image with 60 mm source offset)
left_ref_right_offset

Just recompile the code and run the demo you can also get these results. Do you think this is correct? Because it is hard for me to check all the cuda code, so I'm not sure if there is still minor but import parts I need to consider.

Yifan Deng

### DIFF
--- a/MATLAB/Demos/d025_OffsetSource.m
+++ b/MATLAB/Demos/d025_OffsetSource.m
@@ -22,7 +22,6 @@
 %% Initialize
 clear;
 close all;
-run 'D:\dyf20\CT\TIGRE-offSource\MATLAB\InitTIGRE.m'
 %% Geometry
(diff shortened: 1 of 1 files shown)


AnderBiguri:
I unmerged this, apologies! It seems to cause segmentation faults. If you stillare working on it, do feel free match it to the current TIGRE and try again!

PR #720, opened 2026-02-06
fixed bug in interpolation, added error checks

ReidIsaak:
Fixed bug in interp_weights where invalid weights were returned for input angles outside 360 degrees. Added checks for invalid inputs and outputs.


ReidIsaak:
@AnderBiguri Thanks! Unfortunately, the test data I have available is confidential. Otherwise, I would share it with tests so others could verify.

PR #732, opened 2026-03-11
Fix minor bug and update cc table in setup.py

tsadakane:

PR #738, opened 2026-04-01
fix compute capability

casperdcl:
- fixes #732 (vis. https://developer.nvidia.com/blog/navigating-gpu-architecture-support-a-guide-for-nvidia-cuda-developers/#gpu_support where `nvcc --list-gpu-arch` (on version 12.9) doth not support `compute_110`)
- automatically detect supported architectures
- required to fix builds (e.g. in https://github.com/TomographicImaging/TIGRE-conda/pull/18)


AnderBiguri:
Oh, OK, this is like 1000 times better as a solution!

AnderBiguri:
@casperdcl need a release?

casperdcl:
not personally; happy using just a commit hash for now

Issue #741, opened 2026-04-27
TypeError: only 0-dimensional arrays can be converted to Python scalars


cflin338:
Running the d07_Algorithms02.py demo script, i get the following error. This similar error comes up with later demo scripts too. 

```
SIRT algorithm in progress.
Estimated time until completion : 00:00:00
SART algorithm in progress.
Traceback (most recent call last):
  File "~\TIGRE\Python\demos\d07_Algorithms02.py", line 126, in <module>
    imgSART, qualitySART = algs.sart(
                           ~~~~~~~~~^
        projections,
        ^^^^^^^^^^^^
    ...<7 lines>...
        computel2=True,
        ^^^^^^^^^^^^^^^
    )
    ^
  File "~\.conda\envs\IntroDL\Lib\site-packages\tigre\algorithms\iterative_recon_alg.py", line 475, in iterativereconalg
    alg.run_main_iter()
    ~~~~~~~~~~~~~~~~~^^
  File "~\.conda\envs\IntroDL\Lib\site-packages\tigre\algorithms\iterative_recon_alg.py", line 323, in run_main_iter
    getattr(self, self.dataminimizing)()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "~\.conda\envs\IntroDL\Lib\site-packages\tigre\algorithms\iterative_recon_alg.py", line 346, in art_data_minimizing
    self.update_image(geo, angle, j)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "~\.conda\envs\IntroDL\Lib\site-packages\tigre\algorithms\iterative_recon_alg.py", line 399, in update_image
    * (self.proj[ang_index] - Ax(self.res, geo, angle, "Siddon", gpuids=self.gpuids)),
                              ~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "~\.conda\envs\IntroDL\Lib\site-packages\tigre\utilities\Ax.py", line 59, in Ax
    return _Ax_ext(img, geox, geox.angles, projection_type, geox.mode, gpuids=gpuids)
  File "Python/tigre/utilities/cuda_interface/_Ax.pyx", line 41, in _Ax._Ax_ext
  File "Python/tigre/utilities/cuda_interface/_types.pxd", line 103, in tigre.utilities.cuda_interface._types.convert_to_c_geometry
[...]

PR #742, opened 2026-05-07
fix for #741

VedangNadkarni:
I encountered the same issue as mentioned in #741, and resolved it by removing the blocksize == 1 case. When Ax is called [on line 399 of `update_image` in `iterative_rec_alg.py`](https://github.com/CERN/TIGRE/blob/f04b52757b6b3c7cf00dc3f40066e14984a0abb6/Python/tigre/algorithms/iterative_recon_alg.py#L399), `angle = np.array([self.angleblocks[j]], dtype=np.float32)` creates an extra dimension which does not play well with [`c_geom.DSO[i] = p_geometry.DSO[i]`](https://github.com/CERN/TIGRE/blob/f04b52757b6b3c7cf00dc3f40066e14984a0abb6/Python/tigre/utilities/cuda_interface/_types.pxd#L103) which expects a scalar.

This nature of expected number of dimensions also affects for example [`Python/demos/d09_Algorithms04.py`](https://github.com/CERN/TIGRE/blob/f04b52757b6b3c7cf00dc3f40066e14984a0abb6/Python/demos/d09_Algorithms04.py#L207): `    delta=np.array([-0.005]),` which needs to be changed to `    delta=np.array(-0.005, dtype=np.float32),`. I've not made this change, because I don't know where else such problems exist, but I hope that this change fixes this issue.

### DIFF
--- a/Python/tigre/algorithms/iterative_recon_alg.py
+++ b/Python/tigre/algorithms/iterative_recon_alg.py
@@ -328,14 +328,10 @@ class IterativeReconAlg(object):
 
         for j in range(len(self.angleblocks)):
 
-            if self.blocksize == 1:
-                angle = np.array([self.angleblocks[j]], dtype=np.float32)
-                angle_indices = np.array([self.angle_index[j]], dtype=np.int32)
+            angle = self.angleblocks[j]
+            angle_indices = self.angle_index[j]
 
-            else:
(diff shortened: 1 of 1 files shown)


AnderBiguri:
Thanks @VedangNadkarni !

PR #749, opened 2026-06-23
Fix geometry numpy2 single angle

### LINKED ISSUE #744: single angle geometry issues with np 2.4
## Expected Behavior
Foward/backprojection of a single angle

## Actual Behavior
Foward/Backprojecting a single projection fails. After creating the geometery and angles calling `check_geo()` adds the COR member if not already set.

This creates a nested array per projection (as it does for offDetector and other members)
https://github.com/CERN/TIGRE/blob/a07589f4aebde1aee26f94614d37866aaa6b2782/Python/tigre/utilities/geometry.py#L103-L104

However with the COR parameter (not the others), when this array is indexed it fails with:
```
  File "C:\Users\...\CIL\Wrappers\Python\cil\plugins\tigre\ProjectionOperator.py", line 178, in __call_Ax
    return Ax(data, self.tigre_geom, self.tigre_geom.angles,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "Python/tigre/utilities/cuda_interface/_Ax.pyx", line 41, in _Ax._Ax_ext
  File "Python/tigre/utilities/cuda_interface/_types.pxd", line 158, in tigre.utilities.cuda_interface._types.convert_to_c_geometry
TypeError: only 0-dimensional arrays can be converted to Python scalars
```
This was raising a deprecation warning <numpy 2.4, but an exception after.

Not creating this COR as a nested array for a single projection is a fix, but maybe not the one you want!

```

```

## Specifications 

  - MATLAB/python version:
  - OS:
  - CUDA version: 3.1.3

Paramveersingh-S:
## Summary
Fixes #744 

## Problem
When initializing a geometry with a single projection angle (where `angles.shape[0] == 1`), the `__check_and_repmat__` utility incorrectly tiles 1D scalar attributes (`COR`, `DSO`, `DSD`) into 2D arrays of shape `(1, 1)` instead of maintaining their intended 1D shape `(1,)`. 
In NumPy < 2.4, this triggered a deprecation warning. In NumPy >= 2.4, this raises a fatal `TypeError: only 0-dimensional arrays can be converted to Python scalars` when Cython attempts to parse the geometry.

## Solution
Added a conditional guard inside `__check_and_repmat__` to ensure that `COR`, `DSD`, and `DSO` parameters bypass the `(angles.shape[0], 1)` expansion when their original shape is `(1,)`. They are now correctly tiled as 1D arrays matching `angles.shape[0]`.

## Changes
- `Python/tigre/utilities/geometry.py`: Guarded scalar spatial attributes during geometry instantiation.

### DIFF
--- a/Python/tigre/utilities/geometry.py
+++ b/Python/tigre/utilities/geometry.py
@@ -22,6 +22,7 @@ class Geometry(object):
 
 
     def check_geo(self, angles, verbose=False):
+        angles = np.atleast_1d(np.asarray(angles))
         if angles.ndim == 1:
             self.n_proj = angles.shape[0]
             zeros_array = np.zeros((self.n_proj, 1), dtype=np.float32)
@@ -123,8 +124,9 @@ class Geometry(object):
 
         :return: None
         """
+        exclude_from_cast = ["nVoxel", "nDetector", "n_proj", "mode", "filter"]
         for attrib in self.__dict__:
-            if getattr(self, attrib) is not None:
+            if getattr(self, attrib) is not None and attrib not in exclude_from_cast:
                 try:
                     setattr(self, attrib, np.float32(getattr(self, attrib)))
                 except ValueError:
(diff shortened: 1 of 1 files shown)


AnderBiguri:
Thanks! I was about to write this PR today :D

Issue #678, opened 2025-08-13
Geometry utilities cast to single, casts integers to single too


MargaretDuff:
This bit of code: https://github.com/CERN/TIGRE/blob/989615d8d6c33c2b88c060c14c826ff8b83cce6c/Python/tigre/utilities/geometry.py#L119-L130

## Expected Behavior
Expect anything in double precision is cast to single precision

## Actual Behavior
Casts everything to single precision including integers 

## Code to reproduce the problem (If applicable)
This caused an issue when I did 

```python
self.tigre_geom.cast_to_single()

img, qual= sart(
                proj=self.tigre_projections,
                geo=self.tigre_geom,
                angles=self.tigre_angles,
                init=self.tigre_initial,
                niter=self.niter,
                gpuids=self.gpuids,
                **self.kwargs
```
because sart allocates internally based on `nVoxel` in the geometry which now isn't an integer. 

## Specifications 

  - MATLAB/python version: python
  - OS: linux 
  - CUDA version: 12.9

MargaretDuff:
We have it in CIL and can't remember why we use it either :)

PR #750, opened 2026-06-23
fix(python/geometry): Prevent cast_to_single from casting integers to float32 (#678)

Paramveersingh-S:
## Description
Fixes issue #678 where `Geometry.cast_to_single()` forcefully cast integers and lists of integers into `np.float32`, which inadvertently corrupts integer-based attributes and crashes memory allocations in algorithms like `sart`.

### Changes made
- Added a robust dtype check using `np.issubdtype` to gracefully intercept and ignore pure integers and arrays of integers.
- Floats and float arrays are still correctly cast to single precision.
- No changes required to the `exclude_from_cast` hardcoded array, preventing this issue from occurring on any custom integer variables users might attach to the `Geometry` object.

### DIFF
--- a/Python/tigre/utilities/geometry.py
+++ b/Python/tigre/utilities/geometry.py
@@ -120,15 +120,22 @@ class Geometry(object):
 
     def cast_to_single(self):
         """
-        Casts all number values in current instance to single prevision floating point types.
+        Casts all number values in current instance to single precision floating point types.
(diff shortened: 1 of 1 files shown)


AnderBiguri:
ah thanks! I assume this is one of those numpy 2.4 caused issues, which is why I didn't catch it.

PR #753, opened 2026-07-09
fix: correct multi-dimensional indexing for variable angles

kuniaki-takagi:
Correct multi-dimensional indexing for variable angles in plotproj.py, since there was a bug in d05_plottingFunctions.py.


kuniaki-takagi:
Sorry, this pull request was made by mistake. Closing it now.

PR #760, opened 2026-07-30
Fix FBP filter parameter not propagated to deep copy

cslht:
## Bug

In `FBP()` (`Python/tigre/algorithms/single_pass_algorithms.py:216`), the user-supplied `filter` keyword argument is written to `geo.filter` instead of `geox.filter`. Since `geox` (a deep copy of `geo`) is passed to `filtering()`, the user-specified filter is silently ignored, and always falls back to `ram_lak`.

## Fix

`geo.filter = ...` → `geox.filter = ...` (one line, line 216)

`FDK()` is **not affected** — it reassigns `geo = copy.deepcopy(geo)` at line 180 before setting the filter, so `geo` already refers to the deep copy at that point.

## Tests (GPU-free, mock-based)

- `test_filter_kwarg_goes_to_geox_not_geo` — verifies filter kwarg lands on deep copy, not original
- `test_fbp_default_filter_is_none` — verifies default (no filter kwarg) still works


AnderBiguri:
Thanks foir the catch! however, the test file is not needed, too verbose to just test this.

cslht:
Done — removed the mock-based test file as you suggested. The PR now contains only the one-line fix (`geo.filter` → `geox.filter` at line 216). Thanks for the review!

PR #762, opened 2026-08-03
Fix Varian loader direction and XIM parsing

DebadityaHait:
## Summary

- Correct bowtie calibration-file selection for clockwise Varian TrueBeam 2.7 scans.
- Decode uncompressed XIM pixel buffers into shaped signed-integer arrays.

## Real-data validation

Using TIGRE’s documented Varian CBCT resolution-phantom dataset:
https://zenodo.org/records/8045749

- Base: `StartAngle=180.0`, `StopAngle=-20.0` selects `CC` and loads 10 `FilterBowtie_CC_*` files.
- This PR: the same scan selects `CW` and loads 10 `FilterBowtie_CW_*` files.
- The sample calibration XIM files use `compression=1`.

PR #764, opened 2026-08-03
Fix uncompressed XIM pixel array parsing

DebadityaHait:
﻿﻿## Summary
- Decode uncompressed XIM pixel buffers into shaped signed-integer arrays.
- Add a synthetic 2×2 regression test.

## Testing
- `python -m pytest Python/tests/test_varian_xim.py -q`
- `python -m black --check Python/tests/test_varian_xim.py Python/tigre/utilities/io/varian/xim.py`
- `git diff --check`

### DIFF
--- a/Python/tests/test_varian_xim.py
+++ b/Python/tests/test_varian_xim.py
@@ -1,0 +1,31 @@
+import importlib.util
+import struct
+from pathlib import Path
+
+import numpy as np
+
+XIM_PATH = Path(__file__).resolve().parents[1] / "tigre" / "utilities" / "io" / "varian" / "xim.py"
+
+
+def _load_xim():
+    spec = importlib.util.spec_from_file_location("xim", XIM_PATH)
+    xim = importlib.util.module_from_spec(spec)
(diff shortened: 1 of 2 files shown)


AnderBiguri:
Is this something that happens iwth varian data or just a wild guess of an LLM?

DebadityaHait:
﻿﻿Thanks for asking. This is not based on an anonymized production XIM file. I found a publicly posted Varian XIM reader that explicitly supports both compressed and uncompressed XIMs: https://gist.github.com/1328/7da697c71f9c4ef12e1e

Its uncompressed path reads the pixel buffer as little-endian 32-bit integers and reshapes it to `(height, width)`. I updated this PR’s synthetic fixture to mirror that documented layout and made the parser’s uncompressed integer dtypes explicitly little-endian.

So the regression is source-derived rather than based on a clinical sample. If TIGRE intentionally does not need uncompressed XIM support, I understand if you would prefer not to merge it.

DebadityaHait:
Superseded by #762, which now contains both production changes and the real-data validation.

PR #765, opened 2026-08-05
Release CUDA resources on error, and stop exit(1) in the Python binding

yliu88au:
A CUDA error inside the backprojector currently leaves by the side door. cudaCheckErrors() calls mexErrMsgIdAndTxt() from the middle of voxel_backprojection(), and neither binding survives that well:

  * MATLAB: mexErrMsgIdAndTxt() longjmps back to MATLAB. A longjmp runs no cleanup, so the device buffers, page-locked host memory, host registrations, streams and texture objects held at that moment are all leaked - for the rest of the session, since the GPU context outlives the call.

  * Python: mexErrMsgIdAndTxt() calls exit(1) (Common/CUDA/TIGRE_common.cpp), terminating the host process. There is no exception to catch, no traceback, and the interpreter is gone along with any unsaved work. A library should not end its host process.

The mechanism to do this properly already exists but was never wired up: errors.hpp defines CUDA_SUCCESS/ERR_CUDA/..., voxel_backprojection() is declared to return int, and the Cython layer already calls cuda_raise_errors() on the result - TigreCudaCallError's error_list is documented as "indexed against values in errors.hpp". The function simply always returned 0.

This change:

  * adds Common/CUDA/tigre_cleanup.hpp, a small scope guard that runs registered release actions in reverse order on ANY return path;

  * registers a release action beside each acquisition in voxel_backprojection() - host registrations, device buffers (per device, so a failure part-way through still frees the earlier ones), streams, cudaMallocHost buffers, texture objects/arrays and the plain mallocs. The geometry array is zeroed before use so an early exit cannot free uninitialised pointers;

  * makes cudaCheckErrors() return ERR_CUDA after reporting, which is now safe; void helpers use cudaCheckErrorsVoid();

  * removes the exit(1) - the error text is recorded via tigreSetLastError() for the binding to report;

  * reports at each binding boundary instead: Atb_mex.cpp checks the returned status and raises the MATLAB error there, after cleanup has run, and the Cython layer's existing cuda_raise_errors() starts working because non-zero codes now actually flow;

  * keeps the CUDA error state per-call. cudaGetLastError() is sticky until read, so two things were needed to stop one failure making every later call look broken: clear stale state on entry, and discard errors raised by the cleanup itself (it necessarily runs on half-built state - unregistering a buffer that never registered, freeing an allocation that failed). Without this the next call reports the previous call's error at its first check. Found by the added test, not by inspection.

Behaviour is unchanged on the success path, and unchanged from a MATLAB user's point of view on the error path (same message, same failure) except that the resources are released first. No new CUDA APIs are used, so GPU and CUDA version support is unaffected, and the multi-GPU logic (GpuIds, device loops, splitting) is untouched.

Python/tests/test_cuda_error_propagation.py covers the four properties that matter: the success path still works, repeated calls do not drift GPU memory, a failure raises a catchable error instead of exiting, and the process is still usable afterwards. It reads free memory through ctypes against the CUDA runtime TIGRE already links, so it adds no dependency, and skips those checks if the runtime cannot be located.

Only voxel_backprojection() is converted here to keep the diff reviewable; the same pattern applies to the other entry points and can follow.

Tested on:

  Linux workstation, Ubuntu 22.04.5 (kernel 6.8), 2 x Ada-generation GPUs
  (sm_89, 48 GiB each, driver 580.173.02), CUDA 13.0 (V13.0.88), gcc 11.4.0,
  Python 3.12.13.

  Windows 11 workstation, 1 x Ampere-generation GPU (sm_86, 45 GiB, driver
  596.59, WDDM), CUDA 13.3 (V13.3.73), MSVC 14.44.35207, Python 3.12.12.

Both built clean (no new warnings) and passed all four tests in Python/tests/test_cuda_error_propagation.py. The pair covers two operating systems, two host compilers, two CUDA toolkits and two GPU architectures; the multi-GPU host matters in particular, because the per-device release loop only iterates more than once there.

Not exercised: the MATLAB binding. Atb_mex.cpp is changed to check the returned status and raise the MATLAB error at the boundary rather than from inside the CUDA code, but no MATLAB installation was available to run it, so that half is reviewed-by-inspection only.

### DIFF
--- a/Common/CUDA/TIGRE_common.cpp
+++ b/Common/CUDA/TIGRE_common.cpp
@@ -1,8 +1,9 @@
-#if defined(IS_FOR_PYTIGRE)
 #include <stdlib.h>
 #include <stdio.h>
 #include <stdarg.h>
 #include "TIGRE_common.hpp"
+
+#if defined(IS_FOR_PYTIGRE)
 void mexPrintf(const char* format, ...) {
     PRINT_HERE("");
     va_list argpointer;
@@ -11,10 +12,40 @@ void mexPrintf(const char* format, ...) {
     va_end(argpointer);
 }
 void mexErrMsgIdAndTxt(const char* pcTag, const char* pcMsg) {
+    /* Record and return. This used to exit(1), which ended the host Python
+     * process on any CUDA error - no exception, no traceback, nothing the
+     * caller could catch. Callers now return the errors.hpp codes and the
+     * Cython layer raises TigreCudaCallError, so an error is something the
+     * caller can handle rather than the end of the interpreter. */
     PRINT_HERE("%s %s\n", pcTag, pcMsg);
-    exit(1);
+    tigreSetLastError(pcTag, pcMsg);
 }
 void mexWarnMsgIdAndTxt(const char* pcTag, const char* pcMsg) {
     PRINT_HERE("%s %s\n", pcTag, pcMsg);
 }
 #endif  // IS_FOR_PYTIGRE
+
+/* Detail of the last recorded error, shared by both bindings.
+ *
+ * One static buffer suffices: the CUDA entry points are called one at a time
+ * from the host language, and the binding reads the message immediately after
+ * a non-zero return. */
+static char s_lastError[512] = {0};
+
+void tigreSetLastError(const char* pcTag, const char* pcMsg) {
+    const char* tag = pcTag ? pcTag : "";
+    const char* msg = pcMsg ? pcMsg : "";
+#if defined(_MSC_VER)
+    _snprintf_s(s_lastError, sizeof(s_lastError), _TRUNCATE, "%s %s", tag, msg);
+#else
+    snprintf(s_lastError, sizeof(s_lastError), "%s %s", tag, msg);
+#endif
+}
+
+const char* tigreGetLastError(void) {
+    return s_lastError;
+}
+
+void tigreClearLastError(void) {
+    s_lastError[0] = '\0';
+}
--- a/Common/CUDA/TIGRE_common.hpp
+++ b/Common/CUDA/TIGRE_common.hpp
@@ -21,4 +21,27 @@ void mexWarnMsgIdAndTxt(const char* pcTag, const char* pcMsg);
 #include "mex.h"
 #include "tmwtypes.h"
 #endif  // IS_TIGRE_FOR_PYTHON
+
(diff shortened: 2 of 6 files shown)


AnderBiguri:
Amazing PR, thanks a ton!

yliu88au:
UNOFFICIAL

Hi Ander,

There was always an unexplained memory conflict between TIGRE and CUPY when I run my program, so this is the inspiration for me to dig into it. Glad it was solved for me now.

Regards,
Yi

AnderBiguri:
This PR broke MATLAB compilation, sadly.

AnderBiguri:
Happy to try to investigate further how to make it compatible with both

yliu88au:
Please go ahead. I don't have a working version of MATLAB to test now. Many thanks.

yliu88au:
Found the cause, and it is a build-list omission, not the C++ itself.

#765 moved tigreSetLastError() / tigreGetLastError() / tigreClearLastError() out of the IS_FOR_PYTIGRE block of Common/CUDA/TIGRE_common.cpp and made Atb_mex.cpp report through tigreGetLastError(). Before #765 that file was Python-only (its whole body sat inside #if defined(IS_FOR_PYTIGRE)), so MATLAB/Compile.m never compiled it, and it still does not: the two Atb_mex mex(...) lines list voxel_backprojection*.cu and GpuIds.cpp only. The Atb target therefore fails at link time with an unresolved tigreGetLastError. Nothing else in the MATLAB build references the new symbols (checked with a grep over Common/, MATLAB/, Python/).

Fix (two lines): add '../Common/CUDA/TIGRE_common.cpp' to both Atb_mex lines in Compile.m. Under MATLAB the file compiles only the three last-error functions (the mexPrintf/mexErrMsgIdAndTxt stubs stay Python-only). I also gave Atb_mex.cpp a fallback message: under MATLAB the CUDA-side macros still raise through the real mexErrMsgIdAndTxt(), so the buffer is empty when Atb_mex sees a non-CUDA status code (e.g. no capable device), and the error would otherwise be an empty string.

Branch with #765 re-applied on current master plus that fix: liu005/TIGRE fix/cuda-error-propagation-v2 (commit 9a7c4a2). I cannot compile MATLAB here, so this is reasoned from Compile.m, not tested - if you can run Compile on it, I will open it as a PR (or you are welcome to take the two-line change directly).

PR #767, opened 2026-08-16
parkerweight: take the fan angle from the U axis and fix the weight b…

yliu88au:
…roadcast

The fan coordinate was built from sDetector[0]/dDetector[0] -- the V (row) axis in TIGRE's [V, U] detector convention -- and the weights were applied with np.tile(w, [proj.shape[-1], 1]), whose shape only matches proj[ii]'s (V, U) when the detector is square. On square detectors both errors are invisible; on any anisotropic detector the half-fan angle delta is computed from the wrong extent and the in-place multiply raises a broadcast ValueError.

Square-detector behaviour is unchanged (bit-identical weights).

### DIFF
--- a/Python/tigre/utilities/parkerweight.py
+++ b/Python/tigre/utilities/parkerweight.py
@@ -21,9 +21,11 @@ import numpy as np
 
 
 def parkerweight(proj, geo, angles, q):
-    start = -geo.sDetector[0] / 2 + geo.dDetector[0] / 2
-    stop = geo.sDetector[0] / 2 - geo.dDetector[0] / 2
-    step = geo.dDetector[0]
+    # Detector arrays are [V, U]; the fan of a circular scan is along U (index 1). Using
+    # index 0 only worked on square detectors, where the two extents coincide.
+    start = -geo.sDetector[1] / 2 + geo.dDetector[1] / 2
+    stop = geo.sDetector[1] / 2 - geo.dDetector[1] / 2
+    step = geo.dDetector[1]
     alpha = np.arctan(np.arange(start, stop + step, step) / np.mean(geo.DSD))
     alpha = -alpha
     delta = abs(alpha[0] - alpha[-1]) / 2
@@ -55,7 +57,7 @@ def parkerweight(proj, geo, angles, q):
                 (beta - np.pi - 2 * delta - epsilon) / b_subf(-alpha, delta, epsilon, q) + 0.5
             )  # noqa: E501
         )
-        proj[ii] *= np.tile(w, [proj.shape[-1], 1])
+        proj[ii] *= np.tile(w, [proj.shape[1], 1])
 
     return proj
 


yliu88au:
# Fix parkerweight: fan angle taken from the wrong detector axis

## What is wrong

`Python/tigre/utilities/parkerweight.py` builds its fan-angle coordinate from detector axis 0:

```python
start = -geo.sDetector[0] / 2 + geo.dDetector[0] / 2
stop = geo.sDetector[0] / 2 - geo.dDetector[0] / 2
step = geo.dDetector[0]
```

In TIGRE's geometry convention, detector arrays are ordered `[V, U]` (rows, columns), and the
fan direction of a circular cone-beam scan is along **U** — index 1. The weights are then
applied with

```python
proj[ii] *= np.tile(w, [proj.shape[-1], 1])
```

which replicates `w` across rows, i.e. correctly varies the weight along U — but only
*coincidentally* matches the array it multiplies, because `proj[ii]` has shape `(V, U)` and the
tile has shape `(U, len(w))`.

On the square detectors of every TIGRE example (`V == U`, `sDetector[0] == sDetector[1]`) both
errors are invisible. On any anisotropic detector:

1. the half-fan angle `delta` is computed from the **V** extent, so every Parker transition
   band has the wrong width (a detector that is short in V and wide in U under-weights the
   genuinely redundant rays and vice versa), and
2. `np.tile(w, [proj.shape[-1], 1])` has shape `(U, V)` against `proj[ii]`'s `(V, U)`, so the
   in-place multiply raises a broadcast `ValueError` before the wrong weights can even be
   applied.

## Minimal reproduction

```python
import numpy as np, tigre
from tigre.utilities.parkerweight import parkerweight

geo = tigre.geometry(mode="cone", default=True)
geo.nDetector = np.array([128, 384])          # V, U -- anisotropic
geo.sDetector = geo.nDetector * geo.dDetector
angles = np.linspace(0, np.pi + 0.6, 180, dtype=np.float32)
proj = np.ones((len(angles), 128, 384), dtype=np.float32)
parkerweight(proj, geo, angles, 1)            # ValueError: operands could not be broadcast
```

## Fix

Take the fan coordinate from axis 1 and tile over the row count:

```python
start = -geo.sDetector[1] / 2 + geo.dDetector[1] / 2
stop = geo.sDetector[1] / 2 - geo.dDetector[1] / 2
step = geo.dDetector[1]
...
proj[ii] *= np.tile(w, [proj.shape[1], 1])
```

Behaviour on square detectors is unchanged (bit-identical weights); anisotropic detectors get
the correct fan geometry and no longer crash.

## Out of scope, noted for completeness

`parkerweight` also ignores `geo.offDetector`: a laterally offset detector shifts the fan
centre, which asymmetrizes the redundant bands. Handling that correctly interacts with the
Wang offset weighting elsewhere in the FDK chain, so it is deliberately not part of this
minimal fix.

Found while cross-validating TIGRE's short-scan weighting against an independent Parker
implementation on an industrial CBCT system with a non-square panel readout.

PR #769, opened 2026-08-19
FISTA: default hyper to an automatic power-iteration estimate

yliu88au:
The gradient step in FISTA is 1/hyper, and the right hyper is problem-size dependent - the class docstring itself quotes ~2e8 for a 64^3 head phantom but ~2e4 at 512^3. With the fixed 2.0e8 default, FISTA on any realistically-sized problem takes steps orders of magnitude too small and the reconstruction stays frozen near its initialization. On a synthetic cone-beam phantom (64^3, 90 views) where FDK reaches corr(ground truth) = 0.995, default-hyper FISTA (20 iters) reaches 0.74 with voxel values ~4000x too small - easily mistaken for "FISTA doesn't work on my data".

hyper="auto" (new default) estimates it for the actual problem: power iteration on Atb(Ax(.)) with the same operator pair update_image uses ("interpolated"/"matched"), times 2 x 1.05. The factor 2 is because update_image's effective step 1/hyper acts on the least-squares gradient 2 A^T(Ax-b), whose Lipschitz constant is 2 lambda_max(A^T A) - returning the bare power estimate diverges (verified). The 1.05 biases the from-below-converging power estimate to the safe side. Deterministic seed; cost is n_iter=8 forward+back pairs, paid once at construction.

Passing a numeric hyper reproduces the previous behaviour exactly. Same phantom with hyper="auto": corr = 0.9966 in 20 iterations (FDK 0.9948, CGLS 0.9913 on identical data), correct absolute values.

### DIFF
--- a/Python/tigre/algorithms/ista_algorithms.py
+++ b/Python/tigre/algorithms/ista_algorithms.py
@@ -7,6 +7,7 @@ import tigre
 from tigre.algorithms.iterative_recon_alg import IterativeReconAlg
 from tigre.algorithms.iterative_recon_alg import decorator
 from tigre.utilities.im_3d_denoise import im3ddenoise
+from tigre.utilities.power_method import svd_power_method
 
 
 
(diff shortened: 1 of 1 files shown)


AnderBiguri:
@yliu88au Fantastic! the only thing, can we change the call to use the existing function: https://github.com/CERN/TIGRE/blob/master/Python/tigre/utilities/power_method.py ?

yliu88au:
Good suggestion — done in 5f5d828: _estimate_hyper now uses svd_power_method and returns 2 * 1.05 * sigma_max(A)^2. The factor 2 is because the effective step 1/hyper acts on the least-squares gradient 2 Aᵀ(Ax−b), whose Lipschitz constant is 2·λmax(AᵀA) — the bare σ² diverges; 1.05 biases the from-below-converging estimate to the safe side. Verified against the previous direct estimate on a synthetic cone-beam phantom: estimates within 4%, reconstruction identical to 4 decimals.

PR #771, opened 2026-08-31
Add plotSinogram function for MATLAB (#278)

### LINKED ISSUE #278: Add plotSinogram() to MATLAB
Does the same as plotProj but with a transposed input, slicing on a different axis.

aaravkharwar-hub:
Closes #278.

Adds `plotSinogram.m`, the sinogram counterpart to `plotProj.m`. Instead of slicing over the acquisition angle (dim 3) to show U-V detector images, this slices over the detector row (dim 1) to show U-angle sinograms for each row, matching the request in the issue ("does the same as plotProj but with a transposed input, slicing on a different axis").

Mirrors plotProj's existing option-parsing structure (Step, Colormap, Clims, Savegif, Slice) for consistency with the rest of the MATLAB utilities.

Note: I saw an earlier comment on the issue from another contributor expressing interest, but no branch or PR was ever linked, so I'm submitting this in case it's still open. Happy to close in favor of theirs if it's in progress.


aaravkharwar-hub:
Added `MATLAB/Utilities/test_plotSinogram.m` and actually ran it (via GNU Octave, since I don't have a MATLAB license here) rather than just writing untested code. Two things came out of that:

1. Found and fixed a real bug: `ind=double.empty(0,1)` (copied from `plotProj.m`'s existing pattern) relies on MATLAB's classdef static-method syntax, which errors under Octave. Replaced with `ind=[]` — identical behavior, more portable. This affects `plotProj.m` too, though I've left that file alone since it's out of scope for this PR.
2. The test suite exercises every documented option (Step, Colormap incl. custom matrix, Clims, Slice, Savegif path via error cases) plus a correctness check that the plotted CData for a given detector row actually equals `proj(row,:,:)` — i.e. verifies the transposed-axis behavior the issue asked for, not just "doesn't crash."
All 10 checks pass under Octave. I have not been able to verify against real MATLAB in this environment.

AnderBiguri:
test function is nit necesary, can you remoive it?

aaravkharwar-hub:
Done — removed the test file. The `plotSinogram.m` function itself still has the `double.empty(0,1)` -> `[]` portability fix from that testing, since that's an independent correctness fix unrelated to the test file.

PR #773, opened 2026-09-02
Python fixes: gpuids=None default, helical demo offOrigin axis, computeCOR port

yliu88au:
Three small, independent Python-side items, one commit each so they can be
reviewed (or dropped) separately.

## 1. `gpuids=None` means every visible GPU, not no GPU

Every compiled entry point declares `gpuids=None` in its own signature
(`_Ax_ext`, `_Atb_ext`, `minTV`, `AwminTV`, `tvdenoise`, `add_poisson`), and
all of them convert it through one helper, `convert_to_c_gpuids` in
`_gpuUtils.pxd`. That helper turned `None` into a valid pointer holding
`m_iCount = 0` and a NULL device list, and handed it to CUDA to run on zero
devices.

The algorithm classes hide it — they all do
`if self.gpuids is None: self.gpuids = GpuIds()` before calling down, so
`tigre.Ax`/`Atb` and every `algs.*` function accept `None` happily. The
kernels reachable directly do not: `minTV(img, alpha, iters, None)` **ends
the host process outright** — no exception, no message, no traceback — which
makes it read as "TIGRE is broken" rather than as a bad argument.

Fixed in the helper rather than in seven call sites, since all seven already
route through it: `None` now resolves to `GpuIds()`, whose no-argument form
means "all visible GPUs" — the same thing the algorithm classes were doing by
hand. Also added the missing `raise` on the `malloc` check beside it, which
constructed a `MemoryError` and discarded it.

Tests: `Python/tests/test_gpuids_default.py` (5). Note a regression there
does not *fail* the tests — it aborts the interpreter running them; that is
the nature of the bug being guarded.

(`.pxd` change, so modules that cimport it need recompiling.)

## 2. Helical demo: the helix must move along `offOrigin[:, 0]`, not `[:, 2]`

The Python bindings map `offOrigin` as (z, y, x) — matching numpy's volume
index order (`_types.pyx`) — while MATLAB maps (x, y, z).
`d13_HelicalGeometry.py` was ported from MATLAB's `offOrigin(3,:)` by keeping
the index, which under the Python ordering translates the volume along the
BEAM instead of along the rotation axis: no helix at all.

Verified empirically rather than by reading: `offOrigin[0] = +10` moves a
central ball's projection by exactly the v-shift predicted for a +z volume
shift (+53.1 px at this geometry), while `offOrigin[2] = +10` produces no
v-shift (beam axis, invisible at azimuth 0). A comment at the site records
the ordering and the porting trap, since keeping MATLAB indices is exactly
how the bug was written.

## 3. `computeCOR`: Python parity with `MATLAB/Utilities/computeCOR.m`

The MATLAB toolbox has had `computeCOR.m` (modified from SophiaBeads'
`centre_geom.m`; T. Liu, *Optical Engineering*, April 2009) for years; the
Python bindings have no centre-of-rotation estimator. This adds one, checked
against the MATLAB implementation line by line — the arrays are transposed
relative to it throughout (MATLAB holds (detector, angle), Python
(angle, detector)); the module docstring records the mapping.

One deliberate divergence, matching MATLAB's intent rather than a literal
mask translation: MATLAB's `length(find(test > 0))` is the count of valid
pixels, so its score is a **mean over valid conjugate rays**. A boolean-mask
`len()` in Python would be the angle count — constant across candidates —
turning the search into a sum minimisation that rewards candidates whose
reflected rays fall off the detector (every lost pixel drops a positive
term).

Validated by recovery: a COR planted via `geo.COR` at 0 / ±0.5 / +1.5 /
−2.0 mm comes back within 0.1 mm (test tolerance 0.2 mm). NumPy + SciPy
only, no new dependencies. Tests: `Python/tests/test_compute_cor.py` (6).

---

All three verified on Windows (MSVC, CUDA, RTX A6000): full rebuild from
this branch, then the two new test files pass (11 tests).

### DIFF
--- a/Python/demos/d13_HelicalGeometry.py
+++ b/Python/demos/d13_HelicalGeometry.py
@@ -36,9 +36,16 @@ angles = np.hstack([angles, angles, angles])  # loop 3 times
 # Load thorax phantom data
 head = sample_loader.load_head_phantom(geo.nVoxel)
 
-# This makes it helical
+# This makes it helical.
+#
+# NOTE the component ordering: the Python bindings map offOrigin as
+# (z, y, x) - matching numpy's volume index order - so the axial (helix)
+# motion goes in column 0. This differs from MATLAB, whose offOrigin is
+# (x, y, z) and whose version of this demo therefore uses offOrigin(3,:).
+# Porting by keeping the index silently translates the volume along the
+# BEAM instead of along the rotation axis.
 geo.offOrigin = np.zeros((angles.shape[0], 3))
-geo.offOrigin[:, 2] = np.linspace(
+geo.offOrigin[:, 0] = np.linspace(
     -1024 / 2 + 128, 1024 / 2 - 128, angles.shape[0]
 )  # about 256^3 images fit int he detector with this size.
 
--- a/Python/tests/test_compute_cor.py
+++ b/Python/tests/test_compute_cor.py
@@ -1,0 +1,39 @@
+"""computeCOR: does it run, does it recover a known centre of rotation.
+
+The reference is MATLAB/Utilities/computeCOR.m; see the module docstring of
+tigre/utilities/computeCOR.py for the line-by-line comparison and the one
+deliberate divergence (score normalised by the count of valid pixels, which
+is what MATLAB's `length(find(test > 0))` computes).
+"""
+import numpy as np
+import pytest
+
+import tigre
+
+TOL_MM = 0.2
+
+
+def _problem(cor_mm, nang=120):
+    geo = tigre.geometry(mode="cone", nVoxel=np.array([128, 128, 128]), default=True)
+    angles = np.linspace(0, 2 * np.pi, nang, endpoint=False, dtype=np.float32)
+    geo.check_geo(angles)
+    geo.COR = np.ones(nang, dtype=np.float32) * np.float32(cor_mm)
+    z, y, x = np.mgrid[:128, :128, :128].astype(np.float32)
+    ph = np.zeros((128, 128, 128), dtype=np.float32)
+    ph[((x - 70) ** 2 + (y - 60) ** 2 + (z - 64) ** 2) < 26 ** 2] = 1.0
+    ph[((x - 50) ** 2 + (y - 74) ** 2 + (z - 64) ** 2) < 11 ** 2] = 2.0
+    return geo, angles, tigre.Ax(ph, geo, angles, "Siddon")
+
+
+@pytest.mark.parametrize("true_cor", [0.0, 0.5, -0.5, 1.5, -2.0])
+def test_recovers_a_known_cor(true_cor):
+    geo, angles, proj = _problem(true_cor)
+    est = float(tigre.computeCOR(proj, geo, angles))
+    assert abs(est - true_cor) < TOL_MM, f"true {true_cor}, estimated {est:.4f}"
+
+
+def test_descending_angles_give_the_same_answer():
+    geo, angles, proj = _problem(0.75)
+    a = float(tigre.computeCOR(proj, geo, angles))
+    b = float(tigre.computeCOR(proj[::-1], geo, angles[::-1]))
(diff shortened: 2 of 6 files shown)


AnderBiguri:
2- Are you sure? I would be suprised if this was not thoroughyl tested already. Can you show me the two cases?

yliu88au:
Two independent checks on point 2.

**Code.** `Python/tigre/utilities/cuda_interface/_types.pxd` lines 99-101:

```
c_geom.offOrigX[i] = p_geometry.offOrigin[i][2]
c_geom.offOrigY[i] = p_geometry.offOrigin[i][1]
c_geom.offOrigZ[i] = p_geometry.offOrigin[i][0]
```

so the Python array is (z, y, x). MATLAB's `geo.offOrigin(3,:)` is z, and the literal port `offOrigin[:, 2]` sets x, the beam direction at azimuth 0.

**Projection.** 64³ volume with a ball at the centre, one view at angle 0, default geometry (DSD 1536, DSO 1000, 0.8 mm pixels, 128×128 detector). A +10 mm move along the rotation axis should shift the ball by 10 × 1536/1000 / 0.8 = 19.2 detector rows.

```
offOrigin = [0, 0, 0]     centroid (v, u) = (67.38, 67.38)
offOrigin[:, 0] = +10     centroid (v, u) = (86.60, 67.37)   dv = +19.22 rows   <- along the axis
offOrigin[:, 2] = +10     centroid (v, u) = (67.20, 67.20)   dv =  -0.18 rows   <- along the beam; only magnification changes
```

Running `d13_HelicalGeometry.py` as it is, the volume slides towards the source instead of along the axis, so the acquisition is not helical. Happy to add this as a small test if you want it guarded.

yliu88au:
Apologies, my comment above answered the second *item in the description* (the helical demo); I think your "2" is the second *commit*, `gpuids=None`. Here are the two cases, run on a current master build (Windows, RTX A6000).

**Case 1, works: every wrapped path replaces `None` before it reaches the kernel.**

```python
tigre.Ax(img, geo, angles)                              # Ax.py:51  -> if gpuids is None: gpuids = GpuIds()
algs.asd_pocs(proj, geo, angles, niter=2)               # iterative_recon_alg.py:371/376/381 -> same substitution
```

Both return finite volumes. `im_3d_denoise.py:18` and `pytorch_bindings.py:156` do the same. That is why the existing tests never see the problem: nothing they exercise hands `None` to `convert_to_c_gpuids`.

**Case 2, dies: the compiled entry points called directly, with their *own* default.**

```python
from _minTV import minTV
from _AwminTV import AwminTV
img = np.random.default_rng(0).random((32, 32, 32), dtype=np.float32)

minTV(img, 15.0, 5)                 # gpuids defaults to None in _minTV.pyx:39  -> process ends, exit code != 0, no traceback
minTV(img, 15.0, 5, None)           # same
AwminTV(img, 15.0, 5, -0.005, None) # same
minTV(img, 15.0, 5, GpuIds())       # fine, returns a finite volume
```

With `faulthandler` enabled the last thing printed is `Windows fatal exception: int divide by zero`. The chain is:

1. `convert_to_c_gpuids(None)` (`_gpuUtils.pxd:16`) builds a valid struct with `m_iCount = 0`.
2. `GD_TV.cu:263` does detect `deviceCount == 0` and calls `mexErrMsgIdAndTxt`, but in the Python build that function records the message and *returns* (`TIGRE_common.cpp:14`), so execution continues.
3. `GD_TV.cu:303`: `slices_per_split = (image_size[2] + deviceCount - 1) / deviceCount` — integer division by zero, which is fatal at the OS level and takes the interpreter with it. `GD_AwTV.cu:280/319` is identical.

The fix in the helper makes `None` mean `GpuIds()` (all visible GPUs), which is what every wrapper was already doing by hand, so the seven direct entry points get the same behaviour without touching each one. `Python/tests/test_gpuids_default.py` covers the direct calls; as the docstring there notes, a regression aborts the test process rather than failing an assertion.

PR #774, opened 2026-09-03
Title: Krylov methods: float64 norm accumulation, working restarts, and results that were being discarded

yliu88au:
Four commits, each self-contained and independently checkable; together they
make CGLS/LSQR/LSMR/hybrid_LSQR/AB-GMRES/BA-GMRES/hybrid_fLSQR_TV/IRN_TV_CGLS
behave at production sizes. No CUDA or `.pyx` changes — pure Python, no
rebuild. No new dependencies.

## 1. A float32 `np.linalg.norm` cannot measure a production-sized residual

`np.linalg.norm` accumulates the sum of squares in the array's OWN dtype. On a
projection-sized float32 array the running sum saturates against the
increments still being added, and the answer comes back low. Measured against
a chunked float64 reference on N(0, 1e-3) data at a 3072×3072 detector:

| views | elements | `np.linalg.norm` error |
|---|---|---|
| 45 | 4.2e8 | 0.4 % low |
| 180 | 1.7e9 | 3 % low |
| 360 | 3.4e9 | 16 % low |
| 720 | 6.8e9 | **26 % low** |

A 1024³ volume (1.1e9 elements) is already ~1.4 % off. This is not a
reporting detail: every Krylov step size is a RATIO of such norms
(`alpha = gamma / q_norm**2`), and the loss-of-orthogonality test compares
consecutive residual norms — so at production sizes CGLS both steps wrongly
(volume-norm ÷ projection-norm: the two biases do not cancel) and mistakes
accumulator noise for divergence. On a real 720-view / 1024³ dataset CGLS
"exited due to divergence" at iteration 1; with this change it runs to
completion and its mean lands in family with the other iterative methods.

Two things decide severity, both worth knowing: a **sum of squares** is far
worse than a dot of two different vectors (nothing cancels), and the bias is
**BLAS-dependent** (the same sizes under a different BLAS were ~76× more
accurate), so it can look absent on one machine and be 26 % on another.

Fix: `tigre.utilities.im3Dnorm` gains `l2norm()` and `inner()` that accumulate
in float64 and **return float32**, so no caller's dtype changes (Ax/Atb reject
float64). Routed through them: every Krylov norm, `im3DNORM(...,2)` (so the
ASD-POCS family gets it too), `svd_power_method` (this IS FISTA's Lipschitz
constant) and the POCS TV-step cosine.

## 2. The CGLS/LSQR/LSMR restart never restarted

MATLAB's `CGLS.m` breaks out of an INNER loop into an OUTER one that rebuilds
r/p/gamma. The Python port kept only the inner loop, so that `break` left
`run_main_iter` altogether — the method returned after however few iterations
preceded the first restart. Its give-up sentinel also read
`re_init_at_iteration + 1 == i` against an initial 0, which is true at
`i == 1`. One lost step at iteration 1 ended the reconstruction. LSQR and LSMR
had the identical structure. All three now have the two-loop form and a `-1`
sentinel.

hybrid_LSQR, hybrid_fLSQR_TV and AB/BA-GMRES have no restart path at all — their
sentinel could only fire at one iteration and ignored a rise at every other —
so they now simply stop on a genuine rise.

## 3. Three algorithms discarded their result

`decorator()` ignores what `run_main_iter` RETURNS and hands back `self.res`.
Any `return <expr>` that does not first assign `self.res` therefore returns the
INITIAL volume — zeros, or with an FDK warm start an unimproved FDK image that
looks entirely plausible. `ba_gmres` and `hybrid_lsqr` returned all-zero
volumes on every early exit; **`hybrid_fLSQR_TV` never assigned `self.res` on
any path**, so its output was discarded always.

## 4. AB/BA-GMRES orthogonalised with Python's builtin `sum()`

`h[k,i] = sum(qk.ravel()*w[i])` iterates a CT-sized array element by element:
~40× slower than `np.dot`, 800× less accurate. BA_GMRES's basis is
volume-space → ~107 s per coefficient × 210 coefficients at niter 20; on a
real 1024³ problem that was most of a 4.1 h run (41.6 min after this change).

## 5. IRN_TV_CGLS: an operator that was not an adjoint, and no divergence guard

CGLS runs on the stacked operator `[A; √λ·W·D]` and assumes the handed
transpose IS the adjoint. `Lx`/`Ltx` had an inner-product asymmetry of
**6.4e-2**: `Dxx = np.copy(img)` then writing only `[0:-2]` left the last two
slices holding raw image VALUES (so `D(constant) ≠ 0` — a Tikhonov term
smuggled into part of the volume), and `Dᵀ` was off by one at n-2. Plus no
inner divergence guard (plain CGLS has one): the residual rose from step 2 and
compounded 9.7e3 → 6.9e5. The guard compares only WITHIN one outer iteration —
across a reweighting the residuals belong to different weighted problems. On a
64³ phantom (FDK correlation 0.4163): λ=1 → 0.142 before, **0.4264** after;
λ=100 → 0.451 before, **0.4697** after.

## Tests

- `Python/tests/test_krylov_restart_and_norms.py` (no GPU): float64 accumulation
  vs a chunked reference, float32 return type, restart re-entry and sentinel,
  the D/Dᵀ adjoint, the ordering of `sum()` vs `np.dot`.
- `Python/tests/test_krylov_smoke.py` (GPU): every Krylov method runs, returns a
  finite non-zero volume and improves on its warm start.

Verified on Windows (RTX A6000, CUDA 13.3) against current master. Merge
conflicts with #770 are expected to be trivial (#770 changes 3 lines of
IRN_TV_CGLS scalar dtypes; this branch keeps upstream's `np.sqrt(...,
dtype=np.float32)` form for those lines).

### DIFF
--- a/Python/tigre/algorithms/krylov_subspace_algorithms.py
+++ b/Python/tigre/algorithms/krylov_subspace_algorithms.py
@@ -9,6 +9,7 @@ from tigre.algorithms.iterative_recon_alg import IterativeReconAlg
 from tigre.algorithms.iterative_recon_alg import decorator
 from tigre.utilities.Atb import Atb
 from tigre.utilities.Ax import Ax
+from tigre.utilities.im3Dnorm import l2norm
 import tigre.algorithms as algs
 import scipy.sparse.linalg
 
@@ -21,6 +22,10 @@ else:
 
 def _norm(x, ord=2):
     """Euclidean norm of a large float32 array, accumulated in float64.
+
+    Thin wrapper over ``tigre.utilities.im3Dnorm.l2norm`` so that the Krylov
+    algorithms and the ASD-POCS family - which reaches the same accumulator
+    through ``im3DNORM(..., 2)`` - cannot drift apart.
 
     ``np.linalg.norm`` sums squares in the array's OWN dtype. On a
     projection-sized float32 array the running sum saturates against the tiny
@@ -43,17 +48,9 @@ def _norm(x, ord=2):
     The result is handed back as float32 so that no caller's dtype changes:
     these scalars multiply float32 volumes, and Ax/Atb reject float64.
     """
-    flat = np.ravel(x)
-    if flat.dtype != np.float32:
-        return np.linalg.norm(flat, ord)
     if ord not in (2, None):
-        return np.linalg.norm(flat, ord)
-    total = 0.0
-    step = 1 << 24
-    for k in range(0, flat.size, step):
-        chunk = flat[k:k + step].astype(np.float64)
-        total += float(np.dot(chunk, chunk))
-    return np.float32(np.sqrt(total))
+        return np.linalg.norm(np.ravel(x), ord)
+    return l2norm(x)
 
 
 
--- a/Python/tigre/algorithms/krylov_subspace_algorithms.py
+++ b/Python/tigre/algorithms/krylov_subspace_algorithms.py
@@ -19,6 +19,44 @@ else:
     default_timer = time.clock
 
 
+def _norm(x, ord=2):
+    """Euclidean norm of a large float32 array, accumulated in float64.
+
+    ``np.linalg.norm`` sums squares in the array's OWN dtype. On a
+    projection-sized float32 array the running sum saturates against the tiny
+    increments still being added to it, and the answer comes back low - badly.
+    Measured against a chunked float64 reference on N(0, 1e-3) data of the
+    shape these algorithms actually see (3072 x 3072 detector):
+
+        45 views  (4.2e8 elements)   0.4 % low
+        180 views (1.7e9 elements)     3 % low
+        360 views (3.4e9 elements)    16 % low
+        720 views (6.8e9 elements)    26 % low
+
+    That is not a reporting detail. Every Krylov step size here is a RATIO of
+    such norms (``alpha = gamma / q_norm**2``), and the loss-of-orthogonality
+    test compares consecutive residual norms, so at production sizes the
+    algorithm both steps wrongly and mistakes accumulator noise for
+    divergence. A 1024^3 volume (1.1e9 elements) is already in the bad regime.
+
+    Accumulating the squares in float64 costs one pass and no full-size copy.
+    The result is handed back as float32 so that no caller's dtype changes:
+    these scalars multiply float32 volumes, and Ax/Atb reject float64.
+    """
+    flat = np.ravel(x)
+    if flat.dtype != np.float32:
+        return np.linalg.norm(flat, ord)
+    if ord not in (2, None):
+        return np.linalg.norm(flat, ord)
+    total = 0.0
+    step = 1 << 24
+    for k in range(0, flat.size, step):
+        chunk = flat[k:k + step].astype(np.float64)
+        total += float(np.dot(chunk, chunk))
+    return np.float32(np.sqrt(total))
+
+
+
 class CGLS(IterativeReconAlg):  # noqa: D101
     __doc__ = (
         " CGLS solves the CBCT problem using the conjugate gradient least\n"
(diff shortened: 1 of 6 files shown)


Python/tigre/utilities/im3Dnorm.py
```diff
@@ -1,6 +1,71 @@
...
+        45 views  (4.2e8 elements)   0.4 % low
+        180 views (1.7e9 elements)     3 % low
+        360 views (3.4e9 elements)    16 % low
+        720 views (6.8e9 elements)    26 % low
+
+    A 1024^3 volume (1.1e9 elements) is 1.4 % low. These norms are not
+    only reported - they drive step sizes and stopping rules (CGLS's
+    ``alpha = gamma / q_norm**2``; the ASD-POCS data-fit and TV-step control),
+    so at production sizes the algorithms misbehave rather than merely
+    mis-report.
```
AnderBiguri: Im not sure if all of this is good docs, it seems more like the bug descrition

Python/tigre/algorithms/krylov_subspace_algorithms.py
```diff
@@ -19,6 +20,40 @@
     default_timer = time.clock
 
 
+def _norm(x, ord=2):
+    """Euclidean norm of a large float32 array, accumulated in float64.
```
AnderBiguri: This function seems redundant, as its uniquely used for 2 norm, so it can just call l2norm instead

Python/tigre/algorithms/krylov_subspace_algorithms.py
```diff
@@ -33,62 +68,85 @@ def __init__(self, proj, geo, angles, niter, **kwargs):
...
+        INNER loop iterates. This port previously had only the inner loop, so
+        the `break` that MATLAB uses to fall back into the outer loop and
+        RESTART instead left run_main_iter altogether - CGLS returned after
+        however few iterations preceded the first restart, reporting
+        "exited due to divergence". Combined with a sentinel bug in the
+        give-up test (`re_init_at_iteration + 1 == i` against an initial 0,
+        which is true at i == 1 - MATLAB compares `remember == iter` against
+        an initial 0 that its 1-based counter can never hit on the first
+        check), a single lost step at iteration 1 ended the reconstruction.
+        """
```
AnderBiguri: Are you sure? This code seems to change it to a more MATLAB way, but restarting existed in the code

Python/tigre/algorithms/krylov_subspace_algorithms.py
```diff
@@ -33,62 +68,85 @@ def __init__(self, proj, geo, angles, niter, **kwargs):
...
+            restarted = False
+            while i < self.niter:
                 if self.verbose:
-                    print("re-initilization of CGLS called at iteration:" + str(i))
-                if self.re_init_at_iteration + 1 == i or not self.restart:
-                    print("CGLS exited due to divergence.")
-                    return self.res
-                self.re_init_at_iteration=i
-                i=i-1
-                self.initialize_algo()
```
AnderBiguri, line 77: This was restarting, AFAIK

AnderBiguri:
OK, this is a lot! for the future, I would prefer these to come independently, as its easier to review. 

1- Great! this is indeed something we were trying to tackle mathematically, but this helps a lot!
2-Are you sure? or did the AI miss the restarting code in CGLS?
3-Are you sure??? I am quite sure this has been well tested. Can you show me this happening?
4-great!
5-Not sure I understand, can you explain what changed in the code and where/why?

Python/tigre/utilities/im3Dnorm.py
yliu88au: Fair point. Done in 398d3ea: both docstrings now say only what the function does and why the accumulation is float64 / the return float32. The measurements stay in the PR description.

Python/tigre/algorithms/krylov_subspace_algorithms.py
yliu88au: Agreed. Removed in 398d3ea; all call sites use `l2norm` directly.

Python/tigre/algorithms/krylov_subspace_algorithms.py
yliu88au: The re-init *call* is there, but it is followed directly by `break` (line 80 on master), and there is no outer loop for the break to fall into, so the `for` ends and `run_main_iter` returns. The rebuilt r/p/gamma are never used, and `i = i - 1` has no effect on a `range` iterator. MATLAB's `CGLS.m` breaks out of `for ii` into `while iter<niter`, which rebuilds and continues; that outer loop is what the port lost. LSQR and LSMR have the same shape.

Concrete run on current master, 32³ head phantom, 30 views, `cgls(proj, geo, angles, niter=5, verbose=True)`:

```
re-initilization of CGLS called at iteration:3
l2l = [2585.11, 2466.20, 2279.26, 2653.53, 0.0]
```

The fifth iteration is never executed and the returned volume is the reverted iteration-2 state. Same call on this branch:

```
re-initilization of CGLS called at iteration:3
l2l = [2585.11, 2466.20, 2279.26, 2255.83, 2178.17]
```

Python/tigre/algorithms/krylov_subspace_algorithms.py
yliu88au: See the note on line 96 for the break. The change here is the second half of it: `re_init_at_iteration` starts at 0, so `re_init_at_iteration + 1 == i` is true at `i == 1`, and a rise at the first iteration prints "exited due to divergence" regardless of `restart`. MATLAB's `remember == iter` starts at 0 against a 1-based counter, so it can never fire on the first check. Hence the -1.

yliu88au:
Thanks for the review. Taking the points in order; the two style items are already pushed (398d3ea).

**1.** Glad it helps.

**2. Restart.** The re-init code was there, but it was followed directly by `break` with no outer loop, so `run_main_iter` returned as soon as the first rise happened; `i = i - 1` does nothing to a `range` iterator. On current master, 32³ head phantom, 30 views, `cgls(..., niter=5, verbose=True)`:

```
re-initilization of CGLS called at iteration:3
l2l = [2585.11, 2466.20, 2279.26, 2653.53, 0.0]      # master: iteration 4 never runs
l2l = [2585.11, 2466.20, 2279.26, 2255.83, 2178.17]  # this branch: restarts and continues
```

Separately the give-up test `re_init_at_iteration + 1 == i` starts at 0, so it is true at `i == 1`: a rise at the first iteration ends the run even with `restart=True`.

**3. Discarded results.** `decorator()` (iterative_recon_alg.py, `alg.run_main_iter()` then `return alg.getres()`) never reads what `run_main_iter` returns. `hybrid_fLSQR_TV` never writes `self.res` (it does `return x`, lines 855 and 857), so every call returns the initial volume; `ba_gmres` (line 690) and `hybrid_lsqr` (line 289) do the same on their early-exit branch. On current master, 32³ head phantom, 30 views:

```python
out = algs.hybrid_flsqr_tv(proj, geo, angles, niter=5, lmbda=10)
np.abs(out).max()            # 0.0  -- all zeros
out = algs.hybrid_flsqr_tv(proj, geo, angles, niter=5, lmbda=10, init="FDK")
np.array_equal(out, fdk)     # True -- the FDK warm start handed back untouched
```

Calling `alg.run_main_iter()` directly and keeping its return value gives max 0.42, mean 0.25: that is the reconstruction being thrown away. (On NumPy 2 you first hit a `TypeError` from line 779, which builds a float64 vector; this branch fixes that as well.)

**4.** Thanks.

**5. IRN_TV_CGLS**, three changes, all inside the class:

- `Lx` built the difference with `Dxx = np.copy(img)` and then wrote only `[0:-2]`, so the last two slices held raw image values instead of differences. With `u[i] = i**2` along an axis (n = 16) the last four entries of `D u` should be `[-25, -27, -29, 0]`; master gives `[-25, -27, 196, 225]`. So `D(constant) != 0` there and the regulariser penalises intensity in those slices.
- `Ltx` had the mirror problem at index n-2 (`[1:-2]` leaves it as raw `W*y`), so it was not the adjoint of `Lx`: `<L u, y> = 173.60` vs `<u, Lᵀ y> = 183.93` on random inputs, 6 % off. CGLS applied to `[A; sqrt(lmbda) W D]` needs the exact adjoint, and this mismatch is what made the inner residual climb. Both are now `__D__`/`__Dt__` (forward difference and its exact adjoint; asymmetry 2e-16).
- The inner CGLS loop now has the divergence guard plain CGLS has, compared only within one reweighting (consecutive outer iterations solve different weighted problems), and the outer loop stops when a reweighting does not lower the residual.

On a 64³ phantom (FDK correlation 0.416): lmbda = 1 goes 0.142 → 0.426; lmbda = 100 goes 0.451 → 0.470.

On splitting: happy to open these as five PRs (norms, restarts, results, sum→dot, IRN) if that is easier to review; say the word.

PR #779, opened 2026-09-08
Siddon: a missed ray must RETURN, or the projector never comes back

yliu88au:
**What is wrong.** In `Common/CUDA/Siddon_projection.cu`:

```c
if (am>=aM)
    detector[idx]=0;      // no return: execution falls through
```

`am >= aM` means the ray never enters the volume. Without a `return` the kernel continues to

```c
unsigned long Np=(imax-imin+1)+(jmax-jmin+1)+(kmax-kmin+1);
for (unsigned long ii=0; ii<Np; ii++){ ... }
```

with `imin/imax` inverted for that ray. The sum is a negative float; cast to `unsigned long` it wraps to ~1.8e19, and the loop runs effectively forever. No error, no warning, no result. On the GPU it reads as 100 % SM occupancy with an idle memory controller and no PCIe traffic (every iteration re-fetches the same clamped texture address). Two of our production runs sat in it for 71 minutes before anyone compared against the 0.9 s the same call takes when it works.

**When it trips.** Two conditions together, both of which ordinary circular CT avoids by accident:

- both `nDetector` entries odd: the detector origin is `P.y = dU*(-(nDetecU/2.0)+0.5)`, `P.z = dV*((nDetecV/2.0)-0.5)`, so an odd count shifts that axis by half a pixel, and odd on both nudges grazing rays onto the degenerate side;
- the object is GRAZED rather than cleanly hit or cleanly missed. Measured on a 100 x 125 x 40 mm volume with DSO 407 to 514 mm: centred does not hang, displaced by 50 / 100 / 200 mm hangs, displaced by 300 / 433 mm (every ray misses cleanly) does not hang. It is a band, not a threshold.

Off-axis geometry is ordinary in tomosynthesis and laminography and rare in circular CT, which is where the blind spot comes from. A stock centred cone geometry with an odd x odd detector does NOT reproduce it. Verified on two machines and two GPU architectures (A6000 / sm_86, RTX 6000 Ada / sm_89) over 21 measurements, flipping parity in both directions: adding one pixel to either detector axis rescued a hang, removing one induced it.

**Fix.** `return` after zeroing the pixel, in `Siddon_projection.cu` and `Siddon_projection_parallel.cu`. Value-neutral: every configuration that returned before returns bit-identical sums after.

**Scope, stated precisely.** The cone kernel (`Siddon_projection.cu`) is reproduced, fixed and covered by the test. `Siddon_projection_parallel.cu` carries the identical unsigned-wrap hazard and gets the same `return`, but the hang could not be reproduced in it: 26 configurations on a pre-fix build all returned, over both detector parities and `offOrigin` from 5 to 2000 mm. That half is justified by inspection only (it may be protected by having two terms in `Np`, or by the `copysignf(1e11, ...)` guards it already carries). The `ray_interpolated_*` kernels step a fixed sample count and are unaffected. `Siddon_projection.cu` is compiled into both the Python extension and the MATLAB mex (`Compile.m`), so the fix covers both bindings; the file's only binding difference is the `IS_FOR_MATLAB_TIGRE` detector-index transpose.

**Test** (`Python/tests/test_siddon_missing_ray.py`). Each projection runs in a child process under a timeout, so a regression FAILS instead of hanging the suite. Three tests: the odd x odd detector with the object displaced across the band must return (fails on a pre-fix build by timeout, passes fixed); detector parity must not change the projection sum (even x even reference vs the three odd variants, < 1 %); a centred object is unaffected. `N_ANGLES` is pinned at 179: the offsets and per-view DSD/DSO are interpolated across the view count and at 60 views the test passes on a known-broken build.

**Before / after, on current master.** The hang is build-dependent: the same kernel source compiled with nvcc on Linux trips it, the Windows (MSVC + CUDA 13.3) build of the same source does not - which is why a reproducer is hard to hand over and why the test runs under a timeout.

| build | `test_missed_rays_terminate...` | `test_detector_parity...` | `test_centred_object...` |
|---|---|---|---|
| Linux, RTX 6000 Ada (sm_89), master without the fix | **FAIL: did not return within 90 s** | **FAIL: `[1239, 1179]` did not return within 90 s** | pass |
| Linux, RTX 6000 Ada, this branch | pass (15 s for the three tests) | pass | pass |
| Windows, RTX A6000 (sm_86), master without the fix | pass (does not trip) | pass | pass |
| Windows, RTX A6000, this branch | pass | pass | pass |

The Linux "without the fix" row is the fork's build of the identical `Siddon_projection.cu` (byte-identical to current master), GPUs idle, no co-tenants.


AnderBiguri:
Hmm can you explain in more detail why/when this is failing? I don't see this PR causing any negative consequences anyway, but I would like to understand exactly what is going on.

yliu88au:
Sure — and digging in, my code comment turned out to be wrong about the mechanism, so I've force-pushed a correction (7549d41) along with a figure.

The chain, measured by instrumenting the kernel to dump its intermediates for the ray that hangs (proj 53, row 619, col 313 in the test geometry):

The eps guard snaps ray.x = 0.00025 and ray.z = −4e-6 to exactly 0 — the ray lies along −Y to ~1e-6 rad.
source.x = 280.4 is outside nVoxelX = 64, so −source.x/0 and (nVoxelX−source.x)/0 are both −inf. fminf/fmaxf collapse: axm = axM = −inf.
aM = fminf(axM, ayM, azM) = −inf, am = 0.51 — the miss is detected correctly.
Without the return the Z block runs anyway: kmin = ceilf(source.z + aM·ray.z), and (−inf)·0 = NaN.
Np = NaN, and on CUDA (unsigned long)NaN = 9223372036854775808 (2⁶³). That thread iterates ~9e18 times; the kernel never returns.
I'd originally written that a negative Np wraps to ~1.8e19. Measured, CUDA saturates negatives and −inf to 0 — a zero-trip loop. It's specifically NaN.

[Image]

Top: the x–y plane in voxel coordinates with the real captured numbers. The two red arrows pointing the same way are the whole bug — both X-planes lie on one side of the source, so the two quotients share a sign and the min/max collapse. Green is the same ray with the source inside the volume's x-extent: opposite sides, infinities stay apart, harmless. Bottom: an odd detector count puts a pixel exactly on the axis, so a whole row has ray.z = 0.

When. One ray needs all of: two direction components under eps; the source outside the volume along a zeroed axis; and a miss. It's 2 rays in 261 million here, which is why the symptom is so odd — nvidia-smi shows 100% SM with an idle memory controller and no PCIe, because two threads are re-reading one clamped texture address while everything else finished.

Why odd × odd. Both NaN rays sit on row 619 = (1239−1)/2, the centre row. An odd count puts a pixel exactly on the centre line, so the whole row has ray.z = 0 — 211k candidates for the second coincidence. An even count straddles by half a pixel (0.27 voxels, 270× eps) and never snaps. Adding one pixel to either axis fixes it; removing one re-breaks it. The object offset then decides whether the second coincidence lands: centred doesn't hang, ±50/100/200 mm does, ±300/433 mm doesn't — a band, not a threshold.

Scope. Circular CT isn't immune in principle, but off-axis slab geometry with a large detector sweeps ray directions widely enough that the ~1e-6 rad coincidence actually lands — a stock centred cone-beam with an odd×odd detector did not reproduce it. Siddon_projection_parallel.cu has the same pattern and gets the same return, but I couldn't reproduce a hang there (26 configurations); that half is by inspection and the docstring says so.

Verified on A6000/sm_86 and RTX 6000 Ada/sm_89; bit-identical sums for every configuration that previously returned; 184-test downstream suite unchanged. The figure and its generator are a separate commit (a4f9600) — happy to drop the binary if you'd rather not carry it.

PR #781, opened 2026-09-15
CTnoise.add: use MATLAB's default photon count and clamp non-positive readings

### LINKED ISSUE #780: CTnoise.add with default Poisson returns NaN and noise as large as the signal
## Expected Behavior

With `Poisson` left to its default, `tigre.utilities.CTnoise.add` should behave like MATLAB's `addCTnoise.m`: a default of `I0 = 60000` photons (or `max(proj)/5` when the projections exceed that), which gives low noise and finite projections, and readings at or below zero set to `1e-6` before the log (`Im(Im<=0) = 1e-6`).

## Actual Behavior

The Python port sets the default photon count to

    Poisson = np.ceil(np.log2(np.max(np.abs(projections))))  # nextpow2

That is the exponent, not a photon count: for projections peaking at 300 it is 9. The Lambert-Beer intensities are then between 3 and 9 photons, the Poisson plus Gaussian noise is as large as the signal, and some readings land at or below zero, so `-np.log(projections / Poisson)` gives NaN or inf. There is no clamp before the log either, so explicit low photon counts are affected too, e.g. `CTnoise.add(proj, Poisson=5, Gaussian=np.array([0, 2]))`.

```
import numpy as np
from tigre.utilities import CTnoise

u = np.linspace(-1, 1, 64, dtype=np.float32)
proj = np.tile(300 * np.sqrt(np.clip(1 - u**2, 0, None)), (16, 32, 1)).astype(np.float32)
noisy = CTnoise.add(proj)
print(np.isfinite(noisy).all(), np.abs(noisy - proj).mean())
```

This prints `False nan`, and numpy warns `invalid value encountered in log`. In my runs, between 376 and 416 of the 32,768 values were NaN depending on the seed, the finite values went from about -250 to several thousand for a signal that peaks at 300, and the mean absolute error on the finite values was about 140.

## Specifications 

  - MATLAB/python version: Python 3.12, numpy 2.0.2
  - OS: Linux
  - CUDA version: none. I have no NVIDIA GPU, so I could not build `_RandomNumberGenerator`; the numbers above come from running `CTnoise.py` with `add_noise` replaced by a numpy model of `GeneratePoissonAddGaussian` (Poisson of the input plus `normal * sigma + mu`, float32), over six seeds. The exact values will differ with the real curand RNG, but the NaN come from the default photon count, not from the seed.

## AI Usage

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.

- **Type of assistance**: finding the defect, the reproduction script above, and the text of this issue. No repository code is changed by this issue.
- **Scope**: `Python/tigre/utilities/CTnoise.py` — the default `Poisson` photon count and the missing clamp before the log.
- **How it was found**: by comparing the Python port with its MATLAB reference `MATLAB/Utilities/addCTnoise.m`, then running the script above on `master` and on a fixed build.
- **Level of modification**: the numbers reported here were re-run and checked; the wording is mine to answer for.

Dev-next-gen:
Fixes #780

I was comparing `Python/tigre/utilities/CTnoise.py` with `MATLAB/Utilities/addCTnoise.m` and found that the Python default for `Poisson` is `np.ceil(np.log2(np.max(np.abs(projections))))`. That is the exponent of the next power of two, not a photon count. For projections that peak at 300 it gives 9 photons, so the "realistic" noise ends up as large as the signal. On top of that, the Gaussian term drives some readings to zero or below, and `-np.log()` turns those into NaN or inf. MATLAB avoids both problems: it defaults to `I0 = 60000` (or `max(proj)/5` when the projections go above that) and sets `Im(Im<=0) = 1e-6` before taking the log.

This PR does the same in Python:

- the default photon count is now 60000, or `max(projections) / 5` when the projections exceed 60000, following MATLAB;
- readings `<= 0` after the noise is added are set to `1e-6` before the log, which also covers an explicitly low `Poisson`.

Nothing changes when `Poisson` is passed (every demo passes `Poisson=1e5`), except that a reading that used to become NaN or inf now comes out finite. The default value was never documented, and in practice it could not be used, so I don't think anyone depends on it. One thing to note: the help header of `addCTnoise.m` says the default is 1e5, while its code uses 60000. I followed the code; if you prefer 1e5, it is a one-line change.

`Python/tests/test_ctnoise_defaults.py` checks two things: that the default call returns finite projections with small noise, and that `Poisson=5, Gaussian=[0, 2]` returns finite values. The machine I work on has no NVIDIA GPU, so I ran these tests with `_RandomNumberGenerator` replaced by a numpy model of `GeneratePoissonAddGaussian` (Poisson of the input plus `normal * sigma + mu`, float32). On master both tests fail: the default call gives about 400 NaN out of 32,768 values and finite values from about -250 to several thousand for a signal that peaks at 300. With the change, both pass, and the reproduction script from the issue prints `True` with a mean absolute error of about 1.4. In the repo, the tests import the real `tigre.utilities.CTnoise`, like the other tests there.

## AI Usage

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.

- **Type of assistance**: the AI tool found the defect, wrote the fix and the test, and drafted this description.
- **Scope**: `Python/tigre/utilities/CTnoise.py` (default photon count and clamp before the log) and a new test file, `Python/tests/test_ctnoise_defaults.py`.
- **How it was found**: comparing the Python port with `MATLAB/Utilities/addCTnoise.m`, then confirming with a test that fails before the change and passes after it (RNG stubbed with a numpy model of the CUDA kernel, since no NVIDIA GPU was available).
- **Level of modification**: 6 lines added and 1 removed in `CTnoise.py`, plus a 30-line test file.

Nothing this pipeline produces is submitted without human approval. The entire pull request — the code change, the tests and this description — was reviewed by a human before the pipeline was allowed to submit it.

(@Dev-next-gen)


Dev-next-gen:
I said in the description that I could not run these tests against a compiled TIGRE, only against a numpy model of `GeneratePoissonAddGaussian`. I have now run them on a real GPU, so here is the result with the actual `curand` RNG.

Setup: TIGRE built from this branch with nvcc 12.4, running on an NVIDIA GeForce RTX 3070 Laptop GPU (driver 610.88), Python 3.14, numpy 2.5.3. `tigre.utilities.gpu.getGpuNames()` reports the card, so the CUDA path is the one being exercised.

To isolate the change I swapped only `tigre/utilities/CTnoise.py` in the installed package between the two runs — it is pure Python, so nothing was recompiled and the CUDA extension is identical in both.

With `CTnoise.py` from `master`:

```
2 failed
  test_default_poisson_gives_finite_low_noise_projections FAILED
  test_low_photon_count_does_not_produce_nan_or_inf       FAILED
  assert np.isfinite(noisy).all() -> AssertionError

reproduction script from the issue:
  False nan
  385 NaN out of 32,768, finite values from -221.3 to 2558.0,
  mean absolute error 139.43, RuntimeWarning: invalid value encountered in log
```

With `CTnoise.py` from this branch:

```
2 passed

reproduction script from the issue:
  True 1.4446254
  0 NaN, finite values from -3.8 to 306.4, mean absolute error 1.44, no warning
```

The numbers I quoted in the issue from the stubbed RNG hold up on real hardware: about 400 NaN (385 here) and a mean absolute error around 140 (139.43 here).

One aside, unrelated to this PR: importing the package prints `DeprecationWarning: numpy.core.arrayprint is deprecated` from `Python/tigre/utilities/filtering.py:3`, where `dtype_is_implied` is imported but never used. Happy to send that as a separate one-line PR if it is useful.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.

